PR #721, opened 2013-02-11
Supplement to block display and select positions

nmpetkov:
This supplement is to #715 (here I have forgotten create block template), and to #717, now added "Preview" button in block list (see image below).

Image 1

| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets |
| License | 
| Doc PR |

### DIFF
--- a/src/docs/CHANGELOG.md
+++ b/src/docs/CHANGELOG.md
@@ -37,6 +37,7 @@ Features:
 - Added hook to Blocks module to allow for use with Html Block (only).
 - [BC BREAK] DoctrineExtensions Paginator has been removed, use Doctrine ORM paginator 
   instead http://docs.doctrine-project.org/en/latest/tutorials/pagination.html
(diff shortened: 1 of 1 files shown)


craigh:
@nmpetkov - I am refactoring Blocks and intend to remove this feature (preview). Can you tell me why you find it useful? I do not see a reason for it.

cc: @Guite @cmfcmf

nmpetkov:
I am using this feature very often, and it requires small peace of code to implement. The reasons are:
- When testing a block, it is handy and quick way to see the result;
- The content of the block can be accessed with URL, and this provides possibility for example to share on social networks;
- It is possible to make a page with a block - some blocks have unique content, useful to be accessed with separate URL Here is one example in a test site: http://zk13new.cmstory.com/index.php?module=blocks&type=user&func=display&bid=8

PR #2168, opened 2015-01-06
Merge 1.3 into 1.4 + changelog split

cmfcmf:
This merges the 1.3 branch into the 1.4 branch, obviously ignoring most of the commits in 1.3, because they already are in 1.4.
Why?
From now on, we could make all of our `1.3 + 1.4` related changes to `1.3` and then always merge them into `1.4` (Symfony is doing the same, they apply bug fixes to `2.3` and then merge all the way up to `2.7`). This then would only require one PR, even if it changes stuff in `1.3` and `1.4`.

Although this PR looks very confusing, it's doing nothing more than merging `1.3` into `1.4`.
bildschirmfoto vom 2015-01-06 17 19 27

Please let me know what you think.


Guite:
Interesting :smile:

craigh:
@cmfcmf I don't this this concept will work because the two branches are too divergent now.

craigh:
@cmfcmf I think this should be closed. I split the files in my installer PR maybe we can work from that point forward instead

Issue #2057, opened 2014-11-13
Not found error on all pages only home page works.


Kaik:
Hi,

I have installed fresh copy of current 1.4 (locally) and just after install it gave me 
Not Found
The requested URL /home/Kaik/public_html/core2/src/index.php was not found on this server.

Error occurs at every page whenever I m using old urls or new ones (routing) (url is changing when page reload from old to routing ones) Home page works but with profiler error not found
It might be due to my server configuration its localhost with userdir altrought I have other scripts installed like zikula 1.3.9 and they do work.

Kaik:
Ok I have some informations here. First I did not have installed mod_rewrite but problem still occur after installing. Problem occurs when I was working directly on user_dir so /home/username/public_html/core/src and url localhost/~username/core/src 
While setting up virtual host on same directory and same installation everything works perfectly
url core.dev.com directory home/username/public_html/core/src
since this is how it should work in the first place I don't think there is a need to reopen this issue.

PR #2169, opened 2015-01-06
Corrected behaviour if "mod_rewrite" is not enabled.

cmfcmf:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | ---
| Fixed tickets     | ---
| Refs tickets      | :de: http://support.zikula.de/module-CMS_Support_Forum-viewtopic-topic-31411-start-0.htm#pid149344
| License           | MIT
| Doc PR            | ---
| Changelog updated | no

### DIFF
--- a/src/system/Zikula/Module/SettingsModule/SettingsModuleInstaller.php
+++ b/src/system/Zikula/Module/SettingsModule/SettingsModuleInstaller.php
@@ -37,11 +37,11 @@ class SettingsModuleInstaller extends \Zikula_AbstractInstaller
      */
     public function install()
     {
-        // Set up an initial value for a module variable.  Note that all module
+        // Set up an initial value for a module variable. Note that all module
         // variables should be initialised with some value in this way rather
         // than just left blank, this helps the user-side code and means that
         // there doesn't need to be a check to see if the variable is set in
-        // the rest of the code as it always will be
+        // the rest of the code as it always will be.
         System::setVar('debug', '0');
         System::setVar('sitename', $this->__('Site name'));
         System::setVar('slogan', $this->__('Site description'));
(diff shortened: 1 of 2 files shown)


src/system/Zikula/Module/SettingsModule/SettingsModuleInstaller.php
```diff
@@ -74,7 +74,12 @@ public function install()
         System::setVar('shorturls', false);
         System::setVar('shorturlstype', '0');
         System::setVar('shorturlsseparator', '-');
-        System::setVar('shorturlsstripentrypoint', true);
+
+        if (function_exists('apache_get_modules') && in_array('mod_rewrite', apache_get_modules())) {
+            // Only strip entry point if "mod_rewrite" is available.
+            System::setVar('shorturlsstripentrypoint', true);
+        }
+
```
Guite: @craigh something to consider in your new installer as well

cmfcmf: It's a bit more difficult regarding the installer. I will probably take care of it once the Intaller is finished.

craigh:
so what happens if mod rewrite isn't enabled?

Kaik:
I guess #2057 and it did not work last time when I installed zikula (before this merge) I will check if this is solving the issue.

cmfcmf:
> so what happens if mod rewrite isn't enabled?

Bad stuff :boat: 
1. (fixed by this PR) if stripentrypoint is true, the url will never include `index.php`. Furthermore, the .htaccess rewrite rules aren't applied. This causes all Routes to fail (i.e. `example.com/admin/foo` would search for a file called `foo` in a folder called `admin`, which obviously doesn't exist and display 404).
2. If the URL is `http://example.com` (note the missing `index.php`), the startpage will load, but all routes + debug toolbar fail, because it generates routes and links like `example.com/admin/foo` (without `index.php`). Pending fix: Add rules to the .htaccess (to be done in the installer [I will help with that])

craigh:
would it be prudent to *require* mod_rewrite? Or can we (symfony & routes, etc) get by without it?

cmfcmf:
> would it be prudent to require mod_rewrite?

I also thought about that. I don't have enough experience on whether this is always (or in most of the cases) enabled at hosting providers.

FYI: Drupal 8 requires it as well.

PR #2182, opened 2015-01-17
use variable for session name and use setting from SecurityCenterModule

### LINKED ISSUE #2150: Session cookie name is not working
securitycenter/session settings/Session cookie name:

If I change the session cookie name to e.g. _rksid the cookie will be still named _zsid

craigh:
closes #2150 

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2150
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | n/a

### DIFF
--- a/src/app/config/config.yml
+++ b/src/app/config/config.yml
@@ -18,7 +18,7 @@ framework:
     default_locale:  "%locale%"
     trusted_proxies: ~
     session:
-        name: _zsid
+        name: %zikula.session.name%
         handler_id: session.handler.legacy
         storage_id: session.storage.legacy
     fragments:       ~
--- a/src/app/config/dynamic/default.yml
+++ b/src/app/config/dynamic/default.yml
@@ -16,3 +16,5 @@ swiftmailer:
     auth_mode: null
     delivery_address: null
     disable_delivery: false
+parameters:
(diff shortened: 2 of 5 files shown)


cmfcmf:
I thought you were going to use the dynamic config dumper? :-(

craigh:
I sort of am :smile: In the installer refactor, I abstracted most of it to the YamlDumper class and I'm using that to write directly to `custom_parameters.yml`

do you think I should use the dynamic settings instead?

cmfcmf:
> do you think I should use the dynamic settings instead?

Well why not?

craigh:
> Well why not?

this seemed simpler ;-)

cmfcmf:
> this seemed simpler ;-)

Well then we could just dump everything directly to `custom_parametery.yml` and remove the `dynamic_config.yml`, no?

craigh:
I guess so. why are we doing it with the dynamic config then?

cmfcmf:
> I guess so. why are we doing it with the dynamic config then?

I don't remember :disappointed:

Issue #2197, opened 2015-01-22
mbstring.internal_encoding warning


phaidon:
The use of mbstring.internal_encoding cause a  warning on my page:

 ```
Deprecated: ini_set(): 
Use of mbstring.internal_encoding is deprecated in Zikula/src/lib/legacy/Zikula/Core.php on line 30
```
mac os x - php 5.6.2

craigh:
This won't be changed until at least Core 2.0.0

phaidon:
Sounds reasonable.

shefik:
mbstring.internal_encoding string

Warning

This feature has been DEPRECATED as of PHP 5.6.0. Relying on this feature is highly discouraged.
Defines the default internal character encoding.

PHP 5.6 and later users should leave this empty and set default_charset instead.

http://php.net/manual/en/mbstring.configuration.php#ini.mbstring.internal-encoding

craigh:
closed by #2199

PR #2199, opened 2015-01-23
mbstring.internal_encoding - This feature has been deprecated as of PHP 5.6.0.

shefik:
mbstring.internal_encoding

This feature has been deprecated as of PHP 5.6.0. Relying on this feature is highly discouraged. PHP 5.6 and later users should leave this empty and set default_charset instead. Refs. https://github.com/zikula/core/issues/2197.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | yes
| Tests pass?       | yes
| Fixed tickets     | https://github.com/zikula/core/issues/2197
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/jcss.php
+++ b/src/jcss.php
@@ -13,7 +13,19 @@
  */
 
 include 'lib/bootstrap.php';
-ini_set('mbstring.internal_encoding', 'UTF-8');
+
+/**
+ * mbstring.internal_encoding
+ *
+ * This feature has been deprecated as of PHP 5.6.0. Relying on this feature is highly discouraged.
+ * PHP 5.6 and later users should leave this empty and set default_charset instead.
+ *
+ * @link http://php.net/manual/en/mbstring.configuration.php#ini.mbstring.internal-encoding
+ */
+if (version_compare(\PHP_VERSION, '5.6.0', '<')) {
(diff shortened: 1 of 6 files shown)


craigh:
:-1: Removing the use of a deprecated function is not sufficient. why was that function being used? How do we replace the functionality?

Guite:
Maybe it is related to BC? Because in older installations things were not UTF-8 always.

shefik:
@craigh The commit already addresses what is recommended as a replacement. PHP 5.6 and later users should leave this empty and set default_charset instead.

shefik:
Why does it say "The Travis CI build could not complete due to an error"?

craigh:
> PHP 5.6 and later users should leave this empty and set default_charset instead.

I didn't see that. very good then I guess.

Travis fails because there is an optional library that is now in the composer.json file that requires php 5.4, so the 5.3.x builds all fail for travis.

shefik:
Note: even though the PHP docs say to leave it empty, I found that it didn't work when setting it as an empty string. Don't set it at all to make it work.

Issue #1856, opened 2014-07-03
work on eliminating some BC Breaks


craigh:
refs #1853 

There seems to be some possibility to remove BC breaks in the areas that are currently in the documentation. I suggest we work on seeking and implementing BC where possible.

 - [x]  $this->request->files
 - [x]  $this->request->filter
 - [x]  Paginate (Doctrine Extensions)
 - [x]  Events

$this->request->files
----
note: there is still some question as to whether the old ->getFiles() method might be able to maintain BC or to find another method of providing for BC in this case. (refs #1002, #1261) "one could replace the ParameterBag with a proxy class and magic methods for example. The request object is supposed to be created only once in the front controller btw"

$this->request->filter
----
note: There may still be some effort to provide a BC layer here. (refs #1261) "if someone cares enough it might be possible to handle with a proxy"

Paginate (Doctrine Extensions)
----
note: is this a true BC-break?

Events
----
Could the old events be triggered by the new events for BC purposes only?

craigh:
events were added back in #2050

craigh:
@cmfcmf @Guite can either of you explain what the issue is with **Paginate (Doctrine Extensions)** ? What happened, what was required to be changed and why?

Guite:
Doctrine 2.2 introduced a Paginator component:
http://doctrine-orm.readthedocs.org/en/latest/tutorials/pagination.html

In the 1.3 core we included Doctrine 2.1.4
https://github.com/zikula/core/blob/1.3/src/plugins/Doctrine/Plugin.php#L59
so we had to use an extension for this:
https://github.com/beberlei/DoctrineExtensions#paginator

More information with regards to the different implementations can be found here:
https://github.com/Guite/MostGenerator/blob/master/org.zikula.modulestudio.generator/src/org/zikula/modulestudio/generator/cartridges/zclassic/models/Repository.xtend#L203
https://github.com/Guite/MostGenerator/blob/master/org.zikula.modulestudio.generator/src/org/zikula/modulestudio/generator/cartridges/zclassic/models/Repository.xtend#L732

craigh:
@Guite what about including the Extension in 1.4.0 so both options are available?

Guite:
I wouldn't do this.
First the repo isn't maintained:
```
Warning: This repository is not really maintained anymore. The important paginatior and large collections code was moved to the Doctrine2 core. For all the other extensions, be aware that I don't maintain them anymore.
```
Second how many 1.3 modules are using it actually?
[...]

Issue #1002, opened 2013-08-25
[BC-break] $this->request->files->get()


cmfcmf:
`$this->request->files->get()` returns instances of `Symfony\Component\HttpFoundation\File\UploadedFile` now, before it returned plain arrays out of `$_FILES`.

This means the following code throws a fatal error now:

``` php
$file = $request->files->get('test', null);
echo $file['name']; // 'tmp_name', 'size', 'error', ...
```

PR #2214, opened 2015-02-01
add proxy classes to accommodate BC breaks

craigh:
refs #1856 and refs #1002 
fixes BC-breaks

 - add proxy class for UploadableFile & FileBag to fix issue where file parameters were not accessible by ArrayAccess
 - add proxy class for ParameterBag to fix filter method signature so it is both forward and backward compatible.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | #1856, #1002
| License           | MIT
| Doc PR            | -
| Changelog updated | no

todo
 - [x] approval @cmfcmf 
 - [x] approval @Guite
 - [x] update docs

### DIFF
--- a/src/docs/en/dev/UPGRADE-1.4.0.md
+++ b/src/docs/en/dev/UPGRADE-1.4.0.md
@@ -6,12 +6,8 @@ A very strong effort has been made to keep to the standards of [Semantic Version
 that BC must be maintained within each major version (1.0.0 until 2.0.0). Therefore, all Core-1.3.x modules and themes
 *should* continue to work as expected in Core-1.4.0.
 
-However, because of a necessary upgrade of the Symfony library and despite the development team's best efforts, a few
-BC breaks have still occurred:
-
-  1. [Gedmo (Doctrine Extensions)](#gedmo)
-  2. [Paginate (Doctrine Extensions)](#paginate)
-
+However, because of a necessary upgrade of the Symfony library and despite the development team's best efforts, one
+BC break has still occurred:
 
 <a name="gedmo" />
 Gedmo (Doctrine Extensions)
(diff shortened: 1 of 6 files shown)


craigh:
now that I think about this, maybe I have misused the `__get()` method. maybe I should be using `ArrayAccess` instead. thoughts?

craigh:
OK, I've converted it to `ArrayAccess`. not I need ideas for testing this.

craigh:
now I've added a second fix in the same PR and modified the PR description/title. The second deals with the filter method signature in $request->query->filter()

Guite:
Great :+1:

craigh:
@Guite if you haven't will you please look over this code carefully? I don't know and easy way to test it, so a good eval is important.

cmfcmf:
This looks okay as far as I can see.

craigh:
just opened my local site and got this error:

> ContextErrorException in ParameterBag.php line 60:
Runtime Notice: Declaration of Zikula_Bag_ParameterBag::filter() should be compatible with Symfony\Component\HttpFoundation\ParameterBag::filter($key, $default = NULL, $deep = false, $filter = Symfony\Component\HttpFoundation\FILTER_DEFAULT, $options = Array)

Is there any way to override a method without caring about its current signature?

Guite:
No: afaik if you override a method signatures must be equal. One way to bypass this could be using `__call`: http://php.net/manual/en/language.oop5.overloading.php#object.call

craigh:
I don't think `__call()` will work in this case since the method *exists* in the parent class. Maybe I am misunderstanding.

craigh:
I think this should work now.

craigh:
@Guite ok to merge?

craigh:
FULL STOP!

this is not ready

craigh:
ok, I think it is ready again :stuck_out_tongue_winking_eye:

Issue #1641, opened 2014-02-28
[1.3.7] Module with incompatible core_max/core_min gives error


Portugao:
The module does not get listed in the module list.
It should appear with the icon 'incompatible'.

Portugao:
Should be closed.

PR #2220, opened 2015-02-02
[Extensions] Incompatible modules get saved now

Portugao:
Incompatible modules get saved into the database now with state set to
incompatible.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #1641
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | yes

### DIFF
--- a/src/system/Extensions/lib/Extensions/Api/Admin.php
+++ b/src/system/Extensions/lib/Extensions/Api/Admin.php
@@ -656,8 +656,6 @@ class Extensions_Api_Admin extends Zikula_AbstractApi
 
                             DBUtil::deleteObjectByID('hooks', $modinfo['name'], 'tmodule');
                         }
-                    } else {
-
                     }
                 }
                 unset($tables);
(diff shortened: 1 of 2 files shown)


Guite:
Thanks for the patch!
Could you please create another PR for the 1.4 branch?
@craigh @cmfcmf please correct me if this is wrong.

Portugao:
Puh. I was happy to be able to create this PR.
How to do for 1.4?

craigh:
yes, we need PR's for both branches at the same time so they do not get out of sync.

src/system/Extensions/lib/Extensions/Api/Admin.php
```diff
@@ -1434,7 +1453,7 @@ public function updatehooks($args)
 
         // Delete hook regardless
         $where = "WHERE $hookscolumn[smodule] = '" . DataUtil::formatForStore($modinfo['name']) . "'
-                    AND $hookscolumn[tmodule] <> ''";
+        AND $hookscolumn[tmodule] <> ''";
```
craigh: I disagree with this (and following) formatting change

src/system/Extensions/lib/Extensions/Api/Admin.php
```diff
@@ -621,39 +621,43 @@ public function regenerate($args)
...
+                            if ($hooks) {
+                                foreach ($hooks as $hook) {
+                                    $hook['tmodule'] = $dbmodinfo['name'];
+                                    DBUtil::updateObject($hook, 'hooks');
+                                }
                             }
+
+                            DBUtil::deleteObjectByID('hooks', $modinfo['name'], 'tmodule');
                         }
+                    } else {
```
craigh: why have an `else` if there is nothing else to do?

craigh:
most of this looks OK I think. you can make changes locally and push a new commit to the same branch and the changes will be reflected in the PR (if you did not already know this).

craigh:
> yes, we need PR's for both branches at the same time so they do not get out of sync.

@Portugao - check in 1.4. and see if the problem even exists there. I worked on this issue I think a while ago and maybe already fixed it there. please check.

craigh:
for example in 1.4 I have this:

screen shot 2015-02-02 at 8 17 20 am

craigh:
Is that what you would expect?

Portugao:
@craigh Yes.

craigh:
:+1: great! so no need for a 1.4 PR :wink:

Portugao:
I checked too, it works for me in 1.4. Fine.

PR #2222, opened 2015-02-02
fix issue where routes may conflict with directory structure

### LINKED ISSUE #2215: Unable to access Themes admin
These URLs are not accessible:

/themes/admin
/themes/admin/view

craigh:
Add checks for such in customizing area and upgrade. closes #2215

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2215
| Refs tickets      | -
| License           | MIT
| Doc PR            | included
| Changelog updated | n/a

### DIFF
--- a/src/docs/en/routes_and_urls.md
+++ b/src/docs/en/routes_and_urls.md
@@ -1,0 +1,17 @@
+Routes and Custom Urls
+======================
+
+Routes and some urls can be customized by the site admin. For this example we will use the following sample route:
+
+`/theme/admin/config/`
+
+The first part of this route (`/theme`) is called the prefix and is configurable from within the Extensions module.
+Click on the wrench icon in the module's row and change the value of the **Module URL** to your preference. However,
+there are restrictions on this value; you cannot select a value that is the same as any subdirectory within your
+Zikula installation. So, at least the following names are not allowed: `app`, `config`, `docs`, `images`, `javascript`,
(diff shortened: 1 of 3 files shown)


shefik:
@craigh What happens if the URL is changed via the translatable POT files? Is there a check there too?

Guite:
My guess is that no URLs should be present in the pot files, but they should be assigned as argument to the gettext call.

shefik:
Okay, because the URLs are currently in POT files at the moment.

Guite:
Created a separate issue at #2224 for this-

shefik:
@craigh I tested this commit. It throws an exception "InvalidArgumentException", instead of redirecting the user back to the form to correct the issue, when a URL has the same name of an existing directory.

craigh:
> Okay, because the URLs are currently in POT files at the moment.

where?

craigh:
> It throws an exception

yes. this is the expected behavior

shefik:
@craigh Shouldn't the form be shown again to make it user-friendly? In the POT file search for "theme". That is used as the URL.

craigh:
> Shouldn't the form be shown again to make it user-friendly?

probably, but I just copied the way the API handles other invalid entries, which is an exception. This was done throughout the core. Wrongly IMO, but done and so I kept it consistent.

Issue #2238, opened 2015-02-09
[1.4.x] Tabs disappear when clicked


espaan:
Just tried Content (4.2.0) so old style module etc with 1.4.x. Most things seem to work perfectly.
Of course layouts can be converted into col-md-x columns instead of the "hard coded" 62-38/etc% column divisions, but that is a small thing.

However the Tabs that are using {formtabbedpanel __title='Content'} is not all ok. There are 3 tabs, when you click one the 2nd the first dissappears, when you click on the 3rd the 2nd also dissappears, so in the end you're left with only the 3rd tab. 

The tabs is hidden via jQuery call somehow it seems. Not checked in details where it could come from.

craigh:
I noticed this in EL also.

espaan:
Weird, somehow the tab as well as the content div is hidden instead of only the content div not visible.

craigh:
@phaidon - could this have something to do with some of the "magic" you added to help jQuery and bootstrap play well together?

cmfcmf:
Probably related to https://github.com/zikula/core/issues/1190.

craigh:
refs https://github.com/zikula/core/commit/079df47e7c1f536a0d9eea2993ae19768e1f0554

cmfcmf:
Random guess: It does not work because jQuery is loaded after prototype:
> <script type="text/javascript" src="http://zikula.org/javascript/ajax/proto_scriptaculous.combined.min.js"></script>
<script type="text/javascript" src="http://zikula.org/web/jquery/jquery.min.js"></script>

espaan:
And it should :-) http://api.jquery.com/jquery.noconflict/#jQuery-noConflict-removeAll
the noconflict is always adviced to load prototype first and then jQ

craigh:
no it doesn't work.

craigh:
> And it should :-) 

but the "fix" checks for jQuery - and now this happens BEFORE jQuery exists.

craigh:
we need the sorcerer. ping @phaidon PLEASE HELP!

espaan:
Ah right, that makes it more difficult indeed. mmhh. Then jQuery should be loaded first for the fix, but is adviced to load after prototype in noconflict description..

craigh:
I wonder if the fix could be extracted from it's current location and loaded later.

espaan:
In my case jQuery is loaded before prototype, see below. But still the tabs dissappear.

```
[...]

PR #2242, opened 2015-02-09
implement fix for Prototype jQuery conflict

craigh:
from http://softec.lu/site/DevelopersCorner/BootstrapPrototypeConflict refs #2238


cmfcmf:
Well, this disabled Prototype functionality. Older stuff using Prototype might break.

craigh:
working here: http://zikula.org/library/e/awesome-test-module for me

espaan:
The method in prototype was maybe cleaner only only overriding methods that were hiding.

espaan:
That is exactly what I have here as well and it works great

craigh:
@espaan can you check this again. I added a check for undefined. I assume it won't be a problem, but need to make sure.

espaan:
Hi, just tested with your update. Works perfect. Prototype is loaded before jQuery now, so that should works and it does when loading Content for instance that still uses prototype for some stuff.
And also a pure jQuery page in the admin section loads fine without any JS problems.
So :+1:

craigh:
@phaidon - you said you would look at this. I would like to merge this today if possible

@jguille2 - you had a hand in the original commit on this as well. Any opinions on this solution?

@jusuff - any opinions?

PR #2246, opened 2015-02-09
Revert disabling of translation logging.

cmfcmf:
Merge when https://github.com/schmittjoh/JMSI18nRoutingBundle/pull/144 is merged and a new version is tagged.

### DIFF
--- a/src/app/config/config_dev.yml
+++ b/src/app/config/config_dev.yml
@@ -4,10 +4,6 @@ imports:
 framework:
     router:   { resource: "%kernel.root_dir%/config/routing_dev.yml" }
     profiler: { only_exceptions: false }
-    # translator configuration - logging normally defaults to true in dev env
-    # this is a temp fix until issue #141 at schmittjoh/JMSI18nRoutingBundle is fixed
-    translator:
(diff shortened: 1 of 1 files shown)


craigh:
unless we change the composer.son back to dev/master for this vendor, this probably cannot be merged since the changes at vendor will not occur here. thoughts @cmfcmf ?

craigh:
current is `"jms/i18n-routing-bundle": "1.1.*",`

but [fix](https://github.com/schmittjoh/JMSI18nRoutingBundle/pull/144) is not in a release so not available unless we use dev/master.

craigh:
another concern about which version we use - there are two commits since the last release (before the one referenced above) that might matter also:

 - https://github.com/schmittjoh/JMSI18nRoutingBundle/commit/c5ed2599be62dc65d5d0a4a017fb17f4e76b0aa3
 - https://github.com/schmittjoh/JMSI18nRoutingBundle/commit/bcc32f904f8fe841d883839fccd808900b838313

if any of these commits matter then we should change composer to pull dev/master again. but I need @cmfcmf 's opinion on this.

Issue #656, opened 2013-01-06
Uncaught exception Could not parse dsn


ghost:
I keep getting this error message during install. if i perform a local install there are no problems. but when i try at the web server i get it every time. (External host)
I have an idea that it might be that my web server has an underscore _ in both username and database name. This i can't change!

Copy from install

Uncaught exception Could not parse dsn in /customers/3/d/6/flytkjaer.com/httpd.www/lib/vendor/Doctrine/Doctrine/Manager.php line, 420
#0 /customers/3/d/6/flytkjaer.com/httpd.www/lib/vendor/Doctrine/Doctrine/Manager.php(435): Doctrine_Manager->_buildDsnPartsArray('mysql://flytkja...')
#1 /customers/3/d/6/flytkjaer.com/httpd.www/lib/vendor/Doctrine/Doctrine/Manager.php(305): Doctrine_Manager->parseDsn('mysql://flytkja...')
#2 /customers/3/d/6/flytkjaer.com/httpd.www/lib/vendor/Doctrine/Doctrine/Manager.php(267): Doctrine_Manager->openConnection('mysql://flytkja...', 'default')
#3 /customers/3/d/6/flytkjaer.com/httpd.www/lib/EventHandlers/DoctrineConnector.php(73): Doctrine_Manager::connection('mysql://flytkja...', 'default')
#4 /customers/3/d/6/flytkjaer.com/httpd.www/lib/Zikula/EventManager.php(196): DoctrineListener->doctrineInit(Object(Zikula_Event))
#5 /customers/3/d/6/flytkjaer.com/httpd.www/lib/Zikula/EventManager.php(135): Zikula_EventManager->invoke(Array, Object(Zikula_Event))
#6 /customers/3/d/6/flytkjaer.com/httpd.www/install/lib.php(32): Zikula_EventManager->notify(Object(Zikula_Event))
#7 /customers/3/d/6/flytkjaer.com/httpd.www/install.php(20): install(Object(Zikula_Core))
#8 {main}

PR #2248, opened 2015-02-09
Properly escape characters in DSN.

cmfcmf:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | ---
| Fixed tickets     | http://support.zikula.de/index.php?module=CMS_Support_Forum&func=viewtopic&topic=31410
| Refs tickets      | https://github.com/zikula/core/issues/656
| License           | MIT
| Doc PR            | ---
| Changelog updated | no


craigh:
why is this only in the Doctrine1 connection? is this not a concern in other (symfony/Doctrine2) connections?

cmfcmf:
> why is this only in the Doctrine1 connection? is this not a concern in other (symfony/Doctrine2) connections?

I will further look into it.

Issue #2206, opened 2015-01-27
Categories tree bugs


Guite:
Follow-up from #2189: must be reviewed and fine-tuned, there are some minor bugs left.

* the different context menu actions 
* drag n drop

craigh:
problems remaining:
 - [ ] copy node with children
 - [ ] redraw of tree graphic elements on dynamic (ajax) insert

ping @jusuff

craigh:
note to @Guite : this was NOT an easy pick :stuck_out_tongue_winking_eye:

craigh:
closed by #2253

PR #2253, opened 2015-02-16
[WIP] Category fixes

craigh:
DO NOT MERGE YET

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | #2206
| License           | MIT
| Doc PR            | n/a
| Changelog updated | no

refs #2206

ping @jusuff

### DIFF
--- a/src/lib/util/CategoryUtil.php
+++ b/src/lib/util/CategoryUtil.php
@@ -1084,7 +1084,7 @@ class CategoryUtil
         $title[] = __('Leaf') . ": " . ($category['is_leaf'] ? 'Yes' : 'No');
         $title[] = __('Locked') . ": " . ($category['is_locked'] ? 'Yes' : 'No');
 
-        return implode('&lt;br /&gt;', $title);
+        return implode('<br />', $title);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/lib/util/CategoryUtil.php
```diff
@@ -968,6 +968,33 @@ public static function getCategoryTreeJqueryJS($cats, $doReplaceRootCat = true,
         return $tree->getJqueryHtml();
     }
 
+    public static function getJsTreeNodeFromCategory(\Zikula\Module\CategoriesModule\Entity\CategoryEntity $category)
```
cmfcmf: Does this method need to be public?

craigh: I believe it will be useful publicly, yes.

craigh:
@Guite could use your help in testing this. I think it is done.

Guite:
I tested the current branch trying each context action as well as drag n drop. Overall it looks you did a very great job :-)

Here are the only things I've found:

* Editing a category works, but the updated (localised) name is not reflected in the tree properly until the page is reloaded.
* Moving a category between two other ones or into another one results in a 500 error in the `resequence` method: `Error: Call to a member function setSort_value() on a non-object`

PR #2256, opened 2015-02-18
Allow to hook Scribite to mailer test e-mail settings form

nmpetkov:
This allows site administrator to hook Scribite visual editor provider to the form for test mailer settings. It is useful to prepare HTML formatted mails.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | [no]

### DIFF
--- a/src/system/Zikula/Module/MailerModule/MailerModuleInstaller.php
+++ b/src/system/Zikula/Module/MailerModule/MailerModuleInstaller.php
@@ -50,6 +50,7 @@ class MailerModuleInstaller extends \Zikula_AbstractInstaller
             case '1.3.1':
                 $this->setVar('smtpsecuremethod', 'ssl');
             case '1.3.2':
+            case '1.3.3':
                 // clear old modvars
                 // use manual method because getVars() is not available during system upgrade
                 $modVarEntities = $this->entityManager->getRepository('Zikula\Core\Doctrine\Entity\ExtensionVarEntity')->findBy(array('modname' => $this->name));
@@ -95,6 +96,7 @@ class MailerModuleInstaller extends \Zikula_AbstractInstaller
                 unset($config['spool']);
                 $configDumper->setConfiguration('swiftmailer', $config);
             case '1.4.1':
+                HookUtil::unregisterSubscriberBundles($this->version->getHookSubscriberBundles());
                 HookUtil::registerSubscriberBundles($this->version->getHookSubscriberBundles());
             case '1.4.2':
             // future upgrade routines
--- a/src/system/Zikula/Module/MailerModule/MailerModuleInstaller.php
+++ b/src/system/Zikula/Module/MailerModule/MailerModuleInstaller.php
@@ -50,7 +50,6 @@ class MailerModuleInstaller extends \Zikula_AbstractInstaller
             case '1.3.1':
                 $this->setVar('smtpsecuremethod', 'ssl');
             case '1.3.2':
-            case '1.3.3':
                 // clear old modvars
                 // use manual method because getVars() is not available during system upgrade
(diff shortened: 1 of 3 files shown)


src/system/Zikula/Module/MailerModule/MailerModuleInstaller.php
```diff
@@ -92,6 +95,8 @@ public function upgrade($oldversion)
                 unset($config['spool']);
                 $configDumper->setConfiguration('swiftmailer', $config);
             case '1.4.1':
+                HookUtil::registerSubscriberBundles($this->version->getHookSubscriberBundles());
+            case '1.4.2':
```
craigh: in the other PR, the hooks are registered in 1.3.2 - adding them here is confusing because the versions would differ. I'm not sure of a better solution however. @Guite @cmfcmf your opinion?

Guite:
Why not doing it in 1.3.2 for 1.4, too?

nmpetkov:
1.3.2 in Users and 1.4.2 here are module versions, not Zikula.

Guite:
Sure. My question was: why not register the hooks in 1.3.2 (module) both in 1.3 and 1.4 (core).

nmpetkov:
It is already done.

Guite:
No.
In #2257 you target Mailer 1.3.2, while you target Mailer 1.4.1 here.

nmpetkov:
Mailer has different versions in branch Zikula 1.3 and branch Zikula 1.4, so in both places module version is incremented by one.

Guite:
Repeating my question from above:

Why not doing it in 1.3.2 for 1.4, too?

Guite:
You need to do it in 1.3.2 for both 1.3 and 1.4.
Because otherwise it will be executed twice when people upgrade to 1.4 later.

nmpetkov:
You are right, it is possible registerSubscriberBundles to be executed twice.
But module version in Zikula 1.4 is allready incremented for other reason to 1.4.0.

nmpetkov:
Easy solution is PR for Zikula 1.3 to be ignored, and to remain for Zikula 1.4 only.

craigh:
> But module version in Zikula 1.4 is allready incremented for other reason to 1.4.0.

This is the problem that I was referring to ("the versions differ")

Guite:
Imho we can still add this to the old version in the upgrade method, since 1.4 has not been released yet.

nmpetkov:
I made changes to this PR, with a possible solution.

craigh:
I'm not sure that is the solution either. @Guite is saying that it should be at the same `switch` `case` in both branches. I think I agree but I am concerned about making sure it works.

craigh:
ping @nmpetkov - this needs to be updated further asap or will not make it into 1.4.0

nmpetkov:
Ok, I'll delete PR for version Zikula 1.3, and simplify PR for Zikula 1.4 in short time.

nmpetkov:
I closed PR #2257, and removed unwanted two lines from this PR. So, think now it is safe to be merged.

nmpetkov:
And one more point: now is proper moment to add other subscriber hooks, for example hooked before, or after, sending each e-mail. I don't know if is possible to be added later.

craigh:
> And one more point: now is proper moment to add other subscriber hooks, for example hooked before, or after, sending each e-mail. I don't know if is possible to be added later.

I don't understand your point. Are you planning to add something else here?

nmpetkov:
I only asked if to add more subscriber hooks.

PR #2285, opened 2015-02-23
Correcting form alignment. Refs. #2283

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

PR #2286, opened 2015-02-23
Renaming form label, to remove unnecessary white-space that appears at small break-points

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

form-label

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Form/Type/DbCredsType.php
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Form/Type/DbCredsType.php
@@ -33,7 +33,7 @@ class DbCredsType extends AbstractType
                 'choices' => $this->getDbTypes(),
                 'data' => 'mysql'))
             ->add('dbtabletype', 'choice', array(
-                'label' => __('Database table type (MySQL only)'),
+                'label' => __('Storage Engine'),
                 'label_attr' => array('class' => 'col-lg-5'),
                 'choices' => array(
                     'innodb' => __('InnoDB'),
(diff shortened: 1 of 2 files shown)


craigh:
'storage engine' doesn't seem very user friendly. people know what a DB is, but don't know what a storage engine is.

craigh:
this has some same changes as #2285. The PRs should be mutually exclusive please.

shefik:
@craigh MySQL uses the term "Storage Engine" in their documentation, so it's a valid term to use for consistency.

http://dev.mysql.com/doc/refman/5.1/en/storage-engines.html
http://dev.mysql.com/doc/refman/5.5/en/storage-engine-setting.html

craigh:
I believe you :smile: 

but that doesn't make it any less friendly to the "average joe" that is trying to put together a website.

shefik:
@craigh I corrected the commit you referenced at https://github.com/zikula/core/pull/2285

shefik:
@craigh We really need a different name though. It's too long right now, as illustrated in the screenshot I posted. Also, the commit includes JavaScript which now shows/hide the field, based on the selection for "Database Type", so the text "MySQL Only" is not needed at all anymore. Then, that leaves just "Database Table Type", which would display in three lines. All other labels display in up to two lines. Therefore, we need to have a new name that just takes up two lines max (whether it is "Storage Engine" or something else). If the consensus is to not use "Storage Engine", I'm fine with that, since I can just modify the string later myself anyway in the locale file. However, I am more concerned about having a name that only takes up two lines max.

shefik:
Any other feedback on this pull request?

craigh:
I'll look into it again next week.

Guite:
imho this can be merged. any objections @craigh @cmfcmf ?

Issue #2284, opened 2015-02-23
Fatal error: Call to a member function rowCount()


shefik:
If you submit the Installer form without filling out the form, it throws the following error:

"Fatal error: Call to a member function rowCount() on boolean in /lib/Zikula/Bundle/CoreInstallerBundle/Validator/CoreInstallerValidator.php on line 35"

Zikula 1.4.0 build 452

shefik:
The error occurs when you select "MySQL", but not when you select "PostgreSQL".

craigh:
> the Installer form

which? there are 4

cmfcmf:
@craigh See the error message: `CoreInstallerValidator.php line 35`.

craigh:
@shefik what is the state of the database when you receive this error? Did it have tables?

craigh:
http://stackoverflow.com/questions/1451720/equivalent-of-mysql-list-tables-in-php-pdo

this discussion seems to indicate that doing cross-platform database connection validation is quite difficult until we can use Doctrine which isn't available at this point.

shefik:
@craigh Yes, the database was completely empty. Try submitting the form, without entering any values, and the error will occur. Perhaps more form validation can help?

craigh:
> Perhaps more form validation can help?

the problem is that the validations are occurring in the wrong order apparently. There is already validation for NonNull values

Guite:
@shefik can you please retest with #2310 ?

Guite:
Fixed by #2310. @shefik please reopen in case you still experience problems.

PR #2310, opened 2015-03-04
make table determination more error-prone

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2284 
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Validator/CoreInstallerValidator.php
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Validator/CoreInstallerValidator.php
@@ -26,6 +26,12 @@ class CoreInstallerValidator
      */
     public static function validatePdoConnection($object, ExecutionContextInterface $context)
     {
+        if ($object['database_host'] == '' || $object['database_name'] == '' || $object['database_user'] == '') {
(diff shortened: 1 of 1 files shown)


craigh:
the problem may simply be that the `NotNull` validation needs to occur before the custom Validator in the installer checking for a PDO connection.

craigh:
apparently, one cannot control the order of the validators.

craigh:
How about just checking for a null value before proceeding? instead of looking for errors?

Guite:
Added a check for empty credentials (password is not checked since people could use an empty password on purpose, e.g. for testing).

craigh:
is the first stuff you added still needed then?

Guite:
I would keep it, because it makes the functionality a bit more robust.

cmfcmf:
@craigh  It is possible to control the order of the validator, see http://symfony.com/doc/current/book/validation.html#validation-groups and the next paragraphs.

craigh:
I briefly looked at that before and it didn't seem related, but I will take another look I guess.

PR #2312, opened 2015-03-04
try/catch block here not proper 

### LINKED ISSUE #2293: [1.4] Frontpage setting does not work with arguments set
when Content is used as frontpage module with a certain page set it gives a 404 error. 
You set 
type: user
func: view
parameter: pid=1 for instance
This works fine in current 1.3.x, but has issues in 1.4.x.

See also https://github.com/zikula-modules/Content/issues/160

The type and func seem to pass perfectly to Content, but the argument pid=1 is not passed. 
This seems to come from this line
https://github.com/zikula/core/blob/1.4/src/lib/Zikula/Bundle/CoreBundle/EventListener/LegacyRouteListener.php#L96 where the module is called without arguments for newType (not sure what is newType here) and that does not work with arguments that determine the page to show. When I add $arguments as argument on that specific there it works fine.
So to me that seems it is not a Content issue, but legacyrouter-Core related.

craigh:
as `getModule()` method already wraps the call and outputs object or `null`. so check for `null` instead of catching exception. ping @Guite fixes #2293 

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2293
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

would like some more extensive testing of this to make sure it doesn't affect other things. @Guite you were the original author of what existed. Do you remember what you were trying to fix?

### DIFF
--- a/src/lib/Zikula/Bundle/CoreBundle/EventListener/LegacyRouteListener.php
+++ b/src/lib/Zikula/Bundle/CoreBundle/EventListener/LegacyRouteListener.php
@@ -85,14 +85,8 @@ class LegacyRouteListener implements EventSubscriberInterface
             }
 
             // call the requested/homepage module
-            try {
-                ModUtil::getModule($module);
-                $newType = true;
-            } catch (\Exception $e) {
-                $newType = false;
-            }
-
-            if ($newType) {
+            $moduleBundle = ModUtil::getModule($module);
+            if (null !== $moduleBundle) {
                 $return = ModUtil::func($modinfo['name'], $type, $func);
             } else {
                 $return = ModUtil::func($modinfo['name'], $type, $func, $arguments);


src/lib/Zikula/Bundle/CoreBundle/EventListener/LegacyRouteListener.php
```diff
@@ -85,14 +85,8 @@ public function onKernelRequest(GetResponseEvent $event)
...
-            } catch (\Exception $e) {
-                $newType = false;
-            }
-
-            if ($newType) {
+            $moduleBundle = ModUtil::getModule($module);
+            if (null !== $moduleBundle) {
                 $return = ModUtil::func($modinfo['name'], $type, $func);
             } else {
                 $return = ModUtil::func($modinfo['name'], $type, $func, $arguments);
```
Guite: I wonder if this condition makes sense at all.
Why should we only provide `$arguments` if the module bundle does exist?

craigh: I wondered the same, but left it.

Guite: New-style modules receive no arguments, since `$args` in a controller action is deprecated.

PR #2323, opened 2015-03-07
Re-positioning Messages for Symfony Form Validation

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | https://github.com/zikula-modules/Profile/issues/49
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Resources/views/Install/createadmin.html.twig
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Resources/views/Install/createadmin.html.twig
@@ -14,29 +14,29 @@
         <div class="form-group">
             {{ form_label(form.username) }}
             <div class="col-md-7">
-                {{ form_errors(form.username) }}
                 {{ form_widget(form.username) }}
+                {{ form_errors(form.username) }}
             </div>
         </div>
         <div class="form-group">
             {{ form_label(form.password.first) }}
             <div class="col-md-7">
-                {{ form_errors(form.password.first) }}
                 {{ form_widget(form.password.first) }}
+                {{ form_errors(form.password.first) }}
             </div>
         </div>
         <div class="form-group">
             {{ form_label(form.password.second) }}
             <div class="col-md-7">
-                {{ form_errors(form.password.second) }}
                 {{ form_widget(form.password.second) }}
+                {{ form_errors(form.password.second) }}
             </div>
         </div>
         <div class="form-group">
             {{ form_label(form.email) }}
             <div class="col-md-7">
-                {{ form_errors(form.email) }}
                 {{ form_widget(form.email) }}
+                {{ form_errors(form.email) }}
             </div>
         </div>
         <div class="form-group">
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Resources/views/Install/dbcreds.html.twig
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Resources/views/Install/dbcreds.html.twig
@@ -14,43 +14,43 @@
         <div class="form-group">
             {{ form_label(form.database_driver) }}
             <div class="col-lg-7 col-sm-10">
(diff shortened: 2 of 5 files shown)


craigh:
@shefik - please open a core ticket and update your PR header with that (open) ticket. You can simply copy the contents from the profile ticket if you wish.

I'm not sure how I feel about this change yet. ping @cmfcmf @Guite

Guite:
As @craigh pointed out in https://github.com/zikula-modules/Profile/issues/49 : the current ordering is how Symfony does it, too. See http://symfony.com/doc/current/cookbook/form/form_customization.html for example.
@shefik please elaborate with regards to this.

craigh:
Yes, the default behavior for symfony is the current order (label, error, widget). IMO, this should be retained. If not, then EVERY form in the project will need to be customized every time and I don't this this is a wise direction for the project.

Perhaps your concerns about form flow could be addressed in another way?

I vote :-1: on this PR.

shefik:
All browsers append their HTML5 form validation at the bottom of the field, not at the top. It's not a default behavior, when you are the one laying out the form and where elements get placed on the form in the template. The placement is not coming from Symfony. The documentation from Symfony are only examples, not set in stone as-is. If you have multiple validations associated with a field, the field gets pushed down too far down, which means it's no longer in alignment with its label (of which is a key focal point for the user).

shefik:
Also, the Symfony documentation is just placing one validation above the field. The problem is more evident when you have multiple validations occurring on the same field. The documentation doesn't address this potential problematic layout issue, because the documentation is just one example.

Guite:
So you are saying the Symfony documentation is wrong?
Imho we should discuss this with Symfony people. Because independent of the concrete order we should definitely avoid that Zikula does this different than Symfony.

craigh:
It is **absolutely** a default symfony behavior to have the error appear before the widget. You can see this from [this symfony doc page](http://symfony.com/doc/current/cookbook/form/form_customization.html#form-rendering-basics) where if you simply render the field like `{{ form_row(form.age) }}` then the order of the items that will be rendered is **label, error, widget**. Obviously this can be customized if desired.

Whether you *agree* with their default order is another matter entirely, but there is no question what the default *symfony* behavior is. Whether this aligns with HTML5 is a matter of the symfony team to debate.

IMO, zikula must conform to symfony (right or wrong). I need to be able to render my whole form with `{{ form(form) }}` and know that it will look the same as other zikula forms.

craigh:
However, your points about form flow and appearance are **valid**. So, how can we address this in another fashion (without changing symfony default behavior)?

shefik:
I'm not necessarily saying the Symfony documentation is wrong. It's just showing one example. Their reasoning for placing the validation above field might be due to showing an example more suitable for screen readers and accessibility. Therefore, it would make sense to have a screen reader read the validation message first, and then read the field itself. But, this method is potentially not as friendly for non-screen readers, since it visually effects an efficient user experience, especially when multiple validations might exist for a particular field.

However, HTML has improved, and more efficient methods are available for screen readers, such as "ARIA Live Regions".

More Info:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
http://www.deque.com/blog/accessible-client-side-form-validation-html5-wai-aria/

craigh:
> It's just showing one example.

you keep saying that, but you *mean* that their order is one example of *customization*... the **default** behavior can be seen by rendering the form with `{{ form(form) }}` (which renders the entire form with no customizations).

As I said, your concerns are valid, but this PR is not IMO the solution. We need to find another way.

cmfcmf:
I'd like to stick to the Symfony Docs as well. I think you @shefik could open an issue in the symfony(-docs) projects and make your proposal.

Issue #2321, opened 2015-03-05
Confirm registration validation is working properly


craigh:
#1805 was closed without merging
#1804 removed all the old prototype files and didn't replace any code.
https://github.com/zikula/core/commit/2da396f15a64e745b94fb0826bec687becfd9b0d removed same from admin side since all the files were gone anyway

it is unknown if any JS validation is occurring before a user is created.

has anyone replaced the old js validation?

@phaidon @shefik @jusuff @Guite

craigh:
I checked on this a little. I found that PHP also validates the data so there appears to be no danger of double-registering the same username (for example) or other possible data-corruption.

The old `getRegistrationErrorsAction` method still exists in the ajax controller and could be used if needed. but nothing calls it at the moment.

@jusuff did you say you can look into re-activating ajax validation here or should we discuss another solution?

shefik:
Just add Webshim for the HTML5 validation, as a polyfill for browser support.

craigh:
I don't know anything about those technologies. A PR would be welcome!

Guite:
@shefik this is not about simple checks like mandatory fields. This issue is about extended validation, like username may not be already used, password must not be equal username, email address may not be already used, and so on.

shefik:
@Guite Webshim is still needed for browsers such as Safari on Mac, to add support for attributes such as "required". As an example, the attribute "required" works correctly on Chrome on Mac, and form does not submit via HTML5 validation.

https://github.com/aFarkas/webshim

craigh:
@shefik did your PR only add the library? Did you implement any part of the solution? When I asked for a PR, I assumed it would be understood that you would also do that.

shefik:
@craigh Yes, it only added the Webshim library for the moment, as well as some validation for the installer. I can submit another pull request to specifically address the registration/user forms. I've been going through much of the core in progression, so when I get fully past the installer tickets, I can move on to User and Profile mods, and so on.

craigh:
Feel free to submit many smaller PRs instead of one large one at the end. I would rather have smaller commits anyway

cmfcmf:
Make sure *not* to validate any hooks via Ajax, because this breaks certain modules (such as the Captcha module).

craigh:
> Yes, it only added the Webshim library for the moment, as well as some validation for the installer.

I'm not seeing any usage of this in the installer. can you be more specific where it was added?

craigh:
ping @shefik feedback requested ^^^

[...]

PR #2324, opened 2015-03-07
Adding Webshim to add polyfill for HTML5 form validation in browsers

shefik:
Adding Webshim to add polyfill for HTML5 form validation in browsers, such as adding support in Safari on Mac.

This corrects the installer.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | https://github.com/zikula/core/issues/2321
| License           | MIT
| Doc PR            | -
| Changelog updated | no


shefik:
Note: webshim currently not available to install via composer, so it has been added to the directory "/javascript" instead.

craigh:
this is a very 'heavy' commit. 313 files? seems like a lot for what I think it is doing. can you explain more about the benefits of including this? what does it do?

shefik:
@craigh In this particular case, a browser such as Safari on Mac does not fully support the "required" attribute via HTML5 validation. Therefore, forms are able to be submitted without proper client validation. Webshim provides a variety of polyfills. For the installer, only the polyfills for "forms" is loaded. Even though only the polyfill for "forms" is needed for the installer, all other shims from the library are provided in the core, if developers need them for loading other polyfills, as desired.

Once this is pull request merged, other core modules that have forms can make similar use of proper HTML5 validation. So, it's not only a matter of supporting the "required" attribute cross-browser, but we will be able to properly support HTML5 validation, using attributes such as validation via regex and other valid attributes.

craigh:
:+1:

craigh:
ping @phaidon FYI

shefik:
I opened at ticket at the Webshim project, requesting support to be able to install via composer. In the meantime, the library is located at /javascript/js-webshim.

If a package for composer is made available, the library can be installed to /vendor later on.

https://github.com/aFarkas/webshim/issues/491

cmfcmf:
At least the whole dev folder could have been removed, no? This would probably lower the overall size a lot.

shefik:
@cmfcmf Then, the same can be said about all other libraries where we add both the minified and unminified versions. I would think we need to provide both. (But, have sites only link to the minified version for production use.)

PR #2316, opened 2015-03-05
Multilingual site title, description and meta tags

nmpetkov:
With this PR I'm trying to fix a long awaiting multilingual issue.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #777, #519
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | yes

Some pictures from the interface for entering multilingual items. This is of course subject to change, if sombody like to make better interface.

image

image

Please check changed files, to discuss if this is the best way to implement.

PR #2346, opened 2015-03-11
[1.3] Multilingual site title, description and meta tags

nmpetkov:
This PR is to make available this feature in 1.3 branch.

Once it is accepted and merged, a small PR I'll prepare to 1.4 branch, to coordinate versions of Settings module in both branches.

Changes are well tested both in upgrade and new installs of Zikula 1.3 branch.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| 1.4 PR        | - #2316
| Refs tickets  | -
| License       | LGPLv3+
| Doc PR        | -

### DIFF
--- a/src/lib/i18n/ZLanguage.php
+++ b/src/lib/i18n/ZLanguage.php
@@ -156,6 +156,7 @@ class ZLanguage
         $this->detectLanguage();
         $this->validate();
         $this->fixLanguageToSession();
+        ModUtil::setupMultilingual();
         $this->setLocale($this->languageCode);
         $this->bindCoreDomain();
         $this->processErrors();
--- a/src/lib/util/ModUtil.php
+++ b/src/lib/util/ModUtil.php
@@ -120,6 +120,9 @@ class ModUtil
             }
          }
 
+         // Init multilingual variables here, just to have values, even with default site language (page language is set later)
+         self::setupMultilingual();
+
          // Pre-load the module variables array with empty arrays for known modules that
          // do not define any module variables to prevent unnecessary SQL queries to
          // the module_vars table.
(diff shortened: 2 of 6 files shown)


craigh:
I'm concerned about the versioning. ping @Guite 

we ran into this with another PR a few weeks ago, but I cannot remember who submitted it. If this PR changes the version of the module to 2.9.8 then is there also an upgrade routing in the 1.4 branch with the same version? The module should be continuous in versioning and not duplicate anything. This is part of why features are typically not added to bug-fix branches.

craigh:
to further elaborate, see [this line of code in the 1.4 branch of the settings module](https://github.com/zikula/core/blob/1.4/src/system/Zikula/Module/SettingsModule/SettingsModuleInstaller.php#L149). This line will never run if someone upgrades from 1.3.10 to 1.4.0 because they will already be at the higher version.

So this solution cannot work in my opinion.

nmpetkov:
@craigh My idea for this line of code (https://github.com/zikula/core/blob/1.4/src/system/Zikula/Module/SettingsModule/SettingsModuleInstaller.php#L149), is to move in front of ```case``` for version 2.9.8 (next one). This is safe, because will execute on upgrade from branch 1.3.

craigh:
closing due to versioning conflict. users requiring this feature will have to upgrade to 1.4

nmpetkov:
In fact, version change is only required for deleting variables. This is done in upgrade to Zikula 1.4.
To be clear, new PR is sent, #2358.

nmpetkov:
Remark: in  #2358 the versions both in Zikula 1.3 and 1.4 branch are not touched!

PR #2347, opened 2015-03-12
Settings module versioning

nmpetkov:
Related to:
[1.3] Multilingual site title, description and meta tags #2346

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | - #2346
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/system/Zikula/Module/SettingsModule/SettingsModuleInstaller.php
+++ b/src/system/Zikula/Module/SettingsModule/SettingsModuleInstaller.php
@@ -146,9 +146,12 @@ class SettingsModuleInstaller extends \Zikula_AbstractInstaller
         // Upgrade dependent on old version number
         switch ($oldversion) {
             case '2.9.7':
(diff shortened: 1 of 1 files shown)


craigh:
this puts the module out of sync with itself compared to the old version in 1.3. the code is not all in the same sequence. The code that runs at v2.9.9 will run twice if upgrading from 1.3.10

nmpetkov:
Running code twice will not destroy entered multilingual data earlier (to check, second running will be here: https://github.com/zikula/core/blob/1.4/src/system/Zikula/Module/SettingsModule/SettingsModuleInstaller.php#L167).

It is also possible to move this code to ```case '2.9.7':```. If desired, it can be. If this PR is not acceptable from roadmap point of view, then PR can be closed, and the code can be used from those, who want this features soon, by manually change respective files.

craigh:
closing due to versioning conflict mentioned above.

PR #2372, opened 2015-03-26
View plugin moduleheader

nmpetkov:
This is last step in making easier to style module headers at one place - now with overriding moduleheader.tpl in system module Themes.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -119,3 +119,4 @@ Features:
 - Include Zikula/Wizard - a library to assist in multi-stage user interaction (used in installer)
 - Added multilingual support for site name, site description and site meta tags (#2316).
 - Added view plugin {langchange} for switching language, function with shorturls enabled (#2364)
+- Added view plugin {moduleheader} to unify module headers and make styling at one place - moduleheader.tpl (#2372).
--- a/src/lib/legacy/viewplugins/function.moduleheader.php
+++ b/src/lib/legacy/viewplugins/function.moduleheader.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * Zikula Application Framework
+ *
+ * Copyright Zikula Foundation 2011 - Zikula Application Framework
+ * @license GNU/LGPLv3 (or at your option, any later version).
+ * @package      Zikula_System_Modules
+ * @subpackage   Zikula_Admin
+ */
+
+
+/**
+ * Smarty function build module header in user content page.
+ *
+ * {moduleheader}
+ *
+ * Available parameters:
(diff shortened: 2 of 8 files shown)


craigh:
I'm not sure I'm following what you are doing here. Are you changing the way ALL modules should create their admin header? (no, I didn't really read the code that carefully yet). Could you please elaborate more on what this PR does? It looks like you are making major changes, but not to every module - should all of them be changed? Also, if this is major, it requires a changeling update.

craigh:
Also - some documentation in the developer docs would be required.

craigh:
I've just looked at the code a little more carefully. If I am understanding your intention, this PR is very incomplete. You are creating a 'new' way to create and render module headers for both the admin and the user side. If this is the case, then **every** /system module + Profile and Legal must be changed to be in compliance. Additionally, the documentation of the new way would be very important. (this would be added [here](https://github.com/zikula/core/blob/1.4/src/docs/en/dev/UPGRADE-1.4.0.md)).

but maybe I am misunderstanding. I look forward to more information.

Guite:
Maybe this is something for 1.4.1 ?

nmpetkov:
This PR simply move menu bar (```modulelinks``` plugin) and title band (title and eventually module image) at one place: ```moduleheader``` plugin. This is useful and I think obligatory, to be able to control in a easy way styling of module header - in user area, and in admin area. Particularly it is needed for making for example a bootstrap responsive theme, but not only.

In this PR I've handled with all core modules, which use ```{modulelinks}``` plugin. It was turned out, that they are two only: Search, and Users.

In fact changes are no so big, because in 1.4 branch in Users module {modulelinks} is already moved to ```Users/menu.tpl ```(I've just found a little bug, not from this PR, doubled menu bars, in ```passwordreminder.tpl``` and ```verifyregistration.tpl```, and I'm making additional commit for this just now).

And for Profile and Legal modules: There are pending PRs there, and if proposed changes to the core is accepted, then I'll make small PRs to implement new plugin in them.

craigh:
> It was turned out, that they are two only: Search, and Users.

how is this possible? Doesn't every module have an admin header?

nmpetkov:
Admin yes, but user header - only Search and Users modules.

nmpetkov:
This PR allows for a change of style in one place (one tpl file), but does not change the current style in any way.

PR #2375, opened 2015-03-28
Bootstrap theme overrides

nmpetkov:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | - #2372
| License           | MIT
| Doc PR            | -
| Changelog updated | no

Main purpose of this PR is to make example use of overriding introduced with  #2372 (```moduleheader``` plugin and template), but also is proposal for some theme overrides, to make Bootstrap theme in 1.4 more mobile-devices friendly.

I'm not sure what is purpose of Bootstrap theme, included in the package, but if it is to be used on mobile devices, @craigh, @Guite, you can consider merging this, or just discuss what to do.

I'll make some inline notes in commit. here bellow are two screenshots from my mobile phone, to illustrate the matter.

screenshot_2015-03-28-17-08-53

screenshot_2015-03-28-17-10-04

### DIFF
--- a/src/lib/legacy/viewplugins/function.modulelinks.php
+++ b/src/lib/legacy/viewplugins/function.modulelinks.php
@@ -241,7 +241,7 @@ function smarty_function_modulelinks($params, Zikula_View $view)
                 if (!empty($menuitem['icon'])) {
                     $icon = '<span class="fa fa-'.$menuitem['icon'].'"></span> ';
                 }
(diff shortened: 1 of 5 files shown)


craigh:
> I'm not sure what is purpose of Bootstrap theme, included in the package

me either. I was thinking maybe it should be eliminated? If not, then I guess any positive change is welcome. IIRC, it is based on Andreas08?

Issue #2376, opened 2015-03-28
installer does not allow full character set for ADMIN pwd


craigh:
installer does not allow full character set for ADMIN pwd - because the password is written to a yaml file temporarily, it cannot allow certain characters [without being escaped](http://symfony.com/doc/current/components/yaml/yaml_format.html)

> Strings containing any of the following characters must be quoted. Although you can use double quotes, for these characters it is more convenient to use single quotes, which avoids having to escape any backslash \:

> :, {, }, [, ], ,, &, *, #, ?, |, -, <, >, =, !, %, @, \`

so, form values must be quoted before stored in yaml file.

craigh:
for example, if I create the admin with the password `Bp(2%(!u%P%fT^J` then during the install process, that value is temporarily written to the `custom_parameters.yml` file but it is written without escaping it and the `!` character causes read errors.

I cannot figure out a way to force the Symfony Yaml component to quote the string so it is valid. if I add the quotes myself, then it triples or doubles them.

craigh:
ping @Guite ideas here?

Guite:
My idea would be removing `Zikula\Bundle\CoreBundle\YamlDumper` and using `Symfony\Component\Yaml\Dumper` instead.

craigh:
`Zikula\Bundle\CoreBundle\YamlDumper` uses the functionality of `Symfony\Component\Yaml\Dumper` through the `Symfony\Component\Yaml\Yaml` class.

PR #2379, opened 2015-03-29
encode parameters before writing to ensure values are 'safe'

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2376
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxInstallController.php
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxInstallController.php
@@ -231,7 +231,7 @@ class AjaxInstallController extends AbstractController
     private function updateAdmin()
     {
         $em = $this->container->get('doctrine.entitymanager');
-        $params = $this->yamlManager->getParameters();
+        $params = $this->decodeParameters($this->yamlManager->getParameters());
 
         // create the password hash
         $password = \UserUtil::getHashedPassword($params['password'], \UserUtil::getPasswordHashMethodCode(UsersConstant::DEFAULT_HASH_METHOD));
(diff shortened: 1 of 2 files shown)


craigh:
@shefik - this should fix the password problem.

@Guite or @cmfcmf - I would like your opinion on this. I don't know anything about base64 encoding, so don't know if that is a concern. The encoding isn't for anything other than making the string 'safe' so I doubt there is reason for concern, but wanted to know if you we ok with this before merging.

Guite:
Looks fine imho.

shefik:
@craigh I tested the code, and it fails with the following error:

```
Notice: unserialize(): Error at offset 2 of 6 bytes in /Users/shefik/Documents/SourceTree/Gorilla/shefik/Sources/MacroSyntax/src/lib/Zikula/Core/Doctrine/Entity/ExtensionVarEntity.php on line 81
```

The installation process successfully executes the AJAX. When I click the "Finish" button, the error is thrown.

I used this as the admin password:

```
ATB4n!uWh@f$*Se
```

Guite:
Maybe this relates to the admin email address which is saved as modvar, too.

craigh:
@shefik - please post the entire trace.

craigh:
I cannot reproduce the error using password `ATB4n!uWh@f$*Se`. I suspect that you entered a faulty email address or possibly did not use the correct branch?

I'm merging this.

shefik:
@craigh Confirmed. It is working correctly for me now. Installed Zikula 1.4.0 build 515.

Issue #2385, opened 2015-03-31
Duplicate values in module_vars table


nmpetkov:
The logic of ```ModUtil::setVar``` procedure allows to happen duplicate values for module and config variables. This is truth for both Zikula 1.3 and Zikula 1.4 branches.
Examples here:
#2382

Example for 1.3 branch:
```SELECT modname, name, COUNT( * )  FROM module_vars GROUP BY modname, name HAVING COUNT( * ) >1```
![image](https://cloud.githubusercontent.com/assets/628801/6928472/ef2c7068-d7fc-11e4-8520-4a2d34a5790d.png)

Also, logic in ```ModUtil::initCoreVars``` makes last one to override previous, and it makes impossible to edit this variables in the interface of the site.

**Comment by nmpetkov** [comment]
In 1.3 branch even if duplicates exists, they are ALL updated, as seen in this code fragment:
```
$where = "WHERE $cols[modname] = '" . DataUtil::formatForStore($modname) . "'
                         AND $cols[name] = '" . DataUtil::formatForStore($name) . "'";
$res = DBUtil::updateObject($obj, 'module_vars', $where);
```

In 1.4 branch is updated only FIRST ONE, as seen from the code:
```
$entity = $em->getRepository('Zikula\Core\Doctrine\Entity\ExtensionVarEntity')->findOneBy(array('modname' => $modname, 'name' => $name));
$entity->setValue($value);
```
(using ```findOneBy```)

craigh:
The issue in this case is not ensuring all duplicates are updated, but preventing a duplicate from being created in the first place. As I test this code, the condition you refer to is never reached. So, while *later* updates might be fixed by changing the code you reference, it doesn't fix the initial problem.

Issue #2382, opened 2015-03-31
Localisation settings: can not set different default language


nmpetkov:
How to reproduce:
1. Install one or two other languages (other then English).
2. Go to Settings, Localisation settings.
3. Try to change drop-down "Default language to use for this site".

In my case I can't change.

Analysis:
1. I checked in database, and it is changed, as seen in the picture below:
![image](https://cloud.githubusercontent.com/assets/628801/6919998/51d06780-d7c4-11e4-9dd4-b40e3bb5b193.png)
2. When the site is loaded, and I examine the variable, it is 'en':
![image](https://cloud.githubusercontent.com/assets/628801/6920059/ca9f091e-d7c4-11e4-99e0-1dc233e86500.png)
So, somehow the default language is not set in proper way.

System: New install, Zikula 1.4 build 515.

craigh:
the route is being translated but the strings are not being loaded from the translated catalog. This worked properly a week+ ago. grrrr...

craigh:
if I append `?lang=de` to my url, all the strings are properly translated and it appears to work, so this may be more simple than it appears.

nmpetkov:
I tis not connected to routes I think, just config variable ```language_i18n``` is loaded as 'en' in this line:
https://github.com/zikula/core/blob/1.4/src/lib/i18n/ZLanguage.php#L142
But in database the value is 'de' (for example, after trying to change default language - not by URL, but in settings).

craigh:
> https://github.com/zikula/core/blob/1.4/src/lib/i18n/ZLanguage.php#L142

that line loads from the DB and sets `en` as the default if it cannot load the value.

nmpetkov:
As I examine back trace, this is maybe first call to ```System::getVar```.
```
#0  ZLanguage->__construct() called at [/home/cmstory/public_html/zktest/lib/i18n/ZLanguage.php:181]
#1  ZLanguage::getInstance() called at [/home/cmstory/public_html/zktest/lib/i18n/ZLanguage.php:332]
#2  ZLanguage::getLanguageCode() called at [/home/cmstory/public_html/zktest/lib/util/ModUtil.php:151]
#3  ModUtil::setupMultilingual() called at [/home/cmstory/public_html/zktest/lib/util/ModUtil.php:131]
#4  ModUtil::initCoreVars() called at [/home/cmstory/public_html/zktest/lib/legacy/Zikula/Core.php:519]

```

nmpetkov:
[...]

PR #2386, opened 2015-04-01
Prevent duplicates in module_vars table

nmpetkov:
More robust handling with module and config variables when setting them.
See #2385, #2382.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | - #2385
| Refs tickets      | - #2385, #2382
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/util/ModUtil.php
+++ b/src/lib/util/ModUtil.php
@@ -267,8 +267,8 @@ class ModUtil
         }
 
         $em = ServiceUtil::get('doctrine.entitymanager');
-        if (self::hasVar($modname, $name)) {
-            $entities = $em->getRepository('Zikula\Core\Doctrine\Entity\ExtensionVarEntity')->findBy(array('modname' => $modname, 'name' => $name));
+        $entities = $em->getRepository('Zikula\Core\Doctrine\Entity\ExtensionVarEntity')->findBy(array('modname' => $modname, 'name' => $name));
+        if (count($entities) > 0) {
             foreach($entities as $entity) {
                 // possible duplicates exist. update all (refs #2385)
                 $entity->setValue($value);


craigh:
please don't merge this yet.

craigh:
My concern with this PR is that it substantially changes how `ModUtil::setVar()` is done. The `hasVar()` method was added to prevent unneeded DB calls which this removes. Also, looking back at the commit history on this, there was at one time some major bugs which were introduced and then repaired by @rmburkhead in a rather major commit. I would hate to reintroduce anything like that - especially at this late time in the QA process (I'm hoping we can move to RC3 and Final ASAP).

I believe the two commits I made yesterday  (https://github.com/zikula/core/commit/c08ecae84e4e7f241363a865bfbde73381396fdd & https://github.com/zikula/core/commit/3a3a2fb5040dce05891b3582b42f4a08ffaf58b5) resolve the concerns well enough.

I would like @cmfcmf and @Guite opinions on this please.

Guite:
I agree with @craigh here. Don't know though whether we should close this PR or move it 2.0.0?

nmpetkov:
For me "DB calls" reason is strange. It is obvious, that if you what to SET a variable, then that means you what to make a query to database. And setting a variable (or variables) occurs mainly when editing in admin interface.
Also, it is principle in handling unique data in databases, to check them, before inserting them, otherwise, they have to have proper unique indexes.
If now is not the time for this issue, please move it to the future, but this is an issue at all.

craigh:
for the record, [this](https://github.com/zikula/core/commit/a1bce1d7648789b5af02762ac13fb7a7430eca05) is the commit by @rmburkhead that I was referring to.

@nmpetkov , I am somewhat persuaded by your argument. I wonder if, in addition to your changes here if we should simultaneously attempt to remove the duplicates?

As a matter of principal, I agree that setting a module var implies a DB call to update/insert.

nmpetkov:
OK, I'll be offline 4 days.

craigh:
the array should be updated with the new value also.

I may merge this and make the additional changes myself if I find the time.

@Guite - how do you feel about this after thinking about it for a while?

Guite:
I ran a few test selections on some production sites, but couldn't find any duplicates. So I think this is not a critical issue.
Either merge the proposed fix and any additions you imagine; or postpone the whole matter to 1.4.1 for detailing it after the 1.4 release.

Issue #2387, opened 2015-04-01
registration validation is not working properly


craigh:
refs #2381 #2321 

@Guite :
> @jusuff did you experience problems or missing functionality with regards to the registration validation?

@shefik :
> @jusuff Validation is working for me, including the validation if a user name and/or email address already exists. It also checks for the correct password length and required fields. What issue are you experiencing specifically? Also, with what browser?

craigh:
@shefik - I continue to get errors in my JS console like
> Failed to load resource: the server responded with a status of 412 (Precondition Failed) http://127.0.0.1/core.git/src/javascript/js-webshim/minified/shims/combos/15.js

like I mentioned before.

shefik:
Checking...

craigh:
If I force reload enough times, then the library loads without error, but this is obviously not a workable solution for users. It should load everytime.

shefik:
I tested with Zikula 1.4.0 build 517. I am not experiencing any issues with registration and/or validation.

jusuff:
Ok - issues I reported earlier are related to: #2391 
So I suppose nothing wrong with @shefik  work.

craigh:
closed by #2388

PR #2388, opened 2015-04-01
Re-committing Webshim library.

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | https://github.com/zikula/core/issues/2387
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no


shefik:
@craigh Can you test this commit, to see if you still experience the issue with "412 (Precondition Failed)"?

craigh:
Not for several hours. I am at work for a while

shefik:
I tested with Zikula 1.4.0 build 517, as well as this pull request. I don't see any errors in my JavaScript console.

craigh:
> @craigh Can you test this commit, to see if you still experience the issue with "412 (Precondition Failed)"?

yes, I tested just now with your branch and got the same error.
screen shot 2015-04-02 at 8 43 34 am

craigh:
if I force reload three times, the error goes away. then I type 'craig' as my username (a name I know is already taken) and this is the result:

screen shot 2015-04-02 at 8 44 52 am

there is no error message (field is red though) and the sub text disappears?

shefik:
With Zikula 1.4.0 build 529, routing is now broken for AJAX.

This code:

```
Routing.generate('zikulausersmodule_ajax_getregistrationerrors')
```

Now returns this:

```
http://www.example.com/users/users/ajax/getregistrationerrors
```

The module name is being added twice to the URL.

ping @cmfcmf

shefik:
@craigh It is correct behavior for either the "User Name" or "Email Address" fields to turn red, because the validation for those particular fields is triggered onBlur (where it is checked if a user name and/or email address already exists. I have fixed the issue, where the help text was being removed during this validation. To see the specific error message that triggered the field to invalidate, you must attempt to submit the form. You will then see the entire form apply constraint where applicable, advising you how to correct each field. This is the default behavior for HTML5 browsers, where the error message aren't actually viewable until you click the submit handler for the form.

Regarding the error "412 (Precondition Failed)", I am unable to replicate the issue. I have never encountered the error in my environment. Can you suggest how I can replicate it?

shefik:
@craigh Searching online, I came across this:

```
HTTP error 412 "Precondition Failed" is a server response indicating that the client specified preconditions for response in its request, and that the server could not satisfy that request.

This error is usually caused by the software loading the http resource (usually a webbrowser) so it should have little to nothing to do with your network, but is likely a common configuration setting or applet installed into the browser on all your machines. have you tried alternate browsers on the effected pages? either way, since the error is server based, but is caused by the request stream initiated by your browser, there is no good way to universally fix it. you will have to examine your browser software and determine if a plugin or security software is altering your http header. you can confirm the existence of the precondition field in your header using a protocol analyzer like WireShark.
```

http://superuser.com/questions/512518/412-precodition-failed-error-only-occurs-on-certain-networks

So, based on that info, it might be possible that the error is just happening in your environment. As I mentioned earlier, I am not experiencing the issue. The error has never occurred for me.

craigh:
> With Zikula 1.4.0 build 529, routing is now broken for AJAX.

please rebuild routes and delete your cache and see if the problem persists.

cmfcmf:
> With Zikula 1.4.0 build 529, routing is now broken for AJAX.

Whoops, I think I know what's wrong.

src/javascript/js-webshim/dev/shims/form-combat.js
```diff
@@ -136,10 +122,10 @@ webshims.register('form-combat', function($,webshims){
 	find.register = function(elem, data, pluginDescriptor, plugin){
 		var shadow = typeof pluginDescriptor.shadow == 'string' ? data[pluginDescriptor.shadow] : pluginDescriptor.shadow(data, elem);
 		var shadowFocus = typeof pluginDescriptor.shadowFocus == 'string' ? data[pluginDescriptor.shadowFocus] : pluginDescriptor.shadowFocus(data, elem);
+		console.log(data, arguments);
```
craigh: debug code?

craigh:
> Re-committing Webshim library.

why is this PR needed? does it actually fix #2387? @jusuff thought #2387 was caused by #2391

craigh:
@shefik - I've tried to trouble shoot this but the error is consistent and frustrating. The research I have done says this might be a safari-only problem but most of what I've read has to do with ajax requests and how a `POST` might fail but a `GET` works. I tried to test with Chrome and I think it worked, but other problems are preventing proper testing.

It seems your research about a "server" and "network" doesn't apply as this is all occurring on my local machine using MAMP.

I don't know what to do to proceed. @Guite @jusuff It appears this PR can probably be merged, but proper JS validation doesn't appear to work with Safari.

shefik:
@craig But, I am also using MAMP Pro and Safari, and it works without error for me. What version of Mac OS and version of Safari do you use?

craigh:
OSX 10.2.2 and safari 8.0.4

shefik:
OSX 10.2.2? Is that correct? It's a very old version.

craigh:
sorry - 10.10.2

craigh:
just wondering if all this HTML5 and shims and stuff is overkill for such a simple thing. you've added thousands of lines of code (the webshim lib) for checking two fields that were already checked by jQuery.

jusuff:
Tested validation and it doesn't work well for me. I'm getting 7 ajax request when leaving user name field. Need to test it little more.

shefik:
I don't consider it overkill, when it helps to add native support in the browser. The installer is only using the "forms" shims. All other files are for other features, if anyone wants to load shims for their modules. The installer doesn't use all the shims, and they are not loaded.

Validation for the user name and email address fields is currently triggered onBlur. It can be changed to onSubmit, if you want to limit the request.

I am using MAMP PRO 3.1, Mac OS 10.10.2, Safari 8.0.4, still without any issue on Safari, Chrome, Firefox. 

Do you also experience the 412 error here:

http://jsfiddle.net/trixta/VNuct/

craigh:
> Do you also experience the 412 error here:
http://jsfiddle.net/trixta/VNuct/

no, I don't

jusuff:
@shefik sorry for confusion, but issue I've reported above is not valid anymore - check wrong branch :/

shefik:
@craigh There are also demo forms here - http://afarkas.github.io/webshim/demos/demos/forms.html

That is the developer's website.

Let me know if you experience any 412 error there.

craigh:
merging this to address the concern that jusuff found (multiple ajax requests). but still not convinced this is a solution. I still have my 412 error. with no ideas how to solve.

craigh:
> Let me know if you experience any 412 error there.

I do not.

craigh:
so, I just completely wiped my entire core install. trashed all the files. cleared the DB. recloned. composer-update, reinstalled. loaded `/register` and get the same error in my console before doing anything :confused: .
screen shot 2015-04-03 at 5 27 24 pm

shefik:
@craigh Can you download this file:

https://github.com/aFarkas/webshim/archive/1.15.7.zip

Set up a host in your MAMP environment using those files.

Then, go to this demo page:

/demos/demos/forms.html

Those are the same demo files that are active here:

http://afarkas.github.io/webshim/demos/index.html

Since you said it works error-free for you at that site, let's see if the error persists when you install the same files from that site locally.

shefik:
I don't know if this makes a difference or not, but my local installation is using a build from http://ci.zikula.org/job/Zikula_Core-1.4.0/, whereas you said you are building via composer. Webshim is not compiled vis composer though.

craigh:
> let's see if the error persists when you install the same files from that site locally.

I get no errors loading the library as you asked.

craigh:
> I don't know if this makes a difference or not

I just installed build 536 and get the exact same error as installing from github/composer.

craigh:
I installed build 537 at http://soundwebdev.com/b537/ and still get the same error using Safari. I do not see the error with Chrome.

screen shot 2015-04-04 at 6 03 31 pm

shefik:
@craigh You wouldn't see the problem in Chrome anyway, since it doesn't need the polyfill. So, testing should be in Safari.

shefik:
@craigh Why does your URL have ```/index.php/register```? It should just be ```/register```.

shefik:
@craigh I was able to replicate your error now. It looks like it's somehow related to having Zikula installed in a subdirectory. If Zikula is not in a subdirectory, the error does not occur. So, maybe it's somehow due to mod_rewrite?

PR #2402, opened 2015-04-05
Correcting error "Failed to load resource...412 (Precondition Failed)"

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | https://github.com/zikula/core/pull/2388
| License           | MIT
| Doc PR            | -
| Changelog updated | no

/cc @craigh

### DIFF
--- a/src/javascript/js-webshim/dev/polyfiller.init.js
+++ b/src/javascript/js-webshim/dev/polyfiller.init.js
@@ -1,6 +1,7 @@
 (function($) {
     $.webshim.bugs.bustedValidity = !('minLength' in document.createElement('input'));
     $.webshims.setOptions('basePath', Zikula.Config.baseURL+'javascript/js-webshim/dev/shims/');
+    $.webshims.setOptions('debug', true);
     $.webshims.activeLang(Zikula.Config.lang);
     $.webshims.polyfill(Zikula.Config.polyfillFeatures);
 })(jQuery);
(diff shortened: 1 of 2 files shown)


craigh:
this fixes the problem?

shefik:
I don't see the error anymore in the console.

cmfcmf:
But why does it fix the problem?

shefik:
I'm not quite sure. But, when that that option is set, there is no error. Also, the error only seemed to occur when Zikula is installed in a subdirectory. There was no error when it's installed at the document root.

craigh:
> But why does it fix the problem?

that's what I was wondering :stuck_out_tongue_closed_eyes: 

But @shefik is right, it seems to have fixed the problem. Not only is the console clear, but the functionality is there (I works).

craigh:
@shefik please add a comment in the code explaining why it appears to be in debug mode and then we'll merge it.

shefik:
Without that option being set, "debug" is "undefined" by default (according to the Webshim docs). So, maybe it needs to be set?

craigh:
setting it to false however brings the error back

`$.webshims.setOptions('debug', false);` doesn't work.

shefik:
Correct, I did try as FALSE already. Only setting to TRUE works.

shefik:
I think it might be a bug with the webshim library. So, after this particular commit is merged, I'll open up a ticket at the webshim project, so the developer can take a look himself. The "debug" option shouldn't need to be set.

shefik:
@craigh Comment has been added to the code.

craigh:
@shefik - thank you for continuing to work on this and fix it!

shefik:
You're welcome.

Issue #2407, opened 2015-04-11
Unable to Set Custom Locale


shefik:
Custom text is not being loaded for modules. Using the module Dizkus as an example,

I tested custom locale file in the following paths:

```
/config/locale/en/LC_MESSAGES/module_zikuladizkusmodule.po
/config/locale/en/LC_MESSAGES/module_zikuladizkusmodule.mo
```

or

```
/config/locale/en/LC_MESSAGES/module_dizkus.po
/config/locale/en/LC_MESSAGES/module_dizkus.mo
```

However, the custom text is not being loaded. When I go to this URL:

```
/forums
```
Only the following "Search overrides" domains are loaded:

```
systemplugin_imagine
systemplugin_mapstraction
zikula
```

Line 550 of:

```
/lib/i18n/ZLanguage.php
```

Guite:
Fixed by https://github.com/zikula/core/pull/2408

PR #2408, opened 2015-04-11
Correcting Module Domain

shefik:
This corrects the loading of the module domain, where it was pointing to an incorrect path.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | https://github.com/zikula/core/issues/2407
| Refs tickets      | https://github.com/zikula-modules/DizkusModule/issues/274
| License           | MIT
| Doc PR            | -
| Changelog updated | no


cmfcmf:
I'm not sure if I understand correctly. What is wrong with `zikuladizkusmodule` as translation domain? (sure, it is different from what it was for 1.3.x modules, but as it will only affect newly created 1.4.x modules, this isn't so much of a problem).

shefik:
@cmfcmf I updated the code to remove "module_". I have no issue with the domain just being ```zikuladizkusmodule```. If this is a change for 1.4.x, I'm fine with that.

Note: this pull request still fixes the other issue, where the files were not being loaded correctly.

craigh:
nice work tracking this down and fixing it. I think this needs a changelog entry because the method is changed from 1.3. Also - it would be extra good if it could be documented somewhere else also. maybe the wiki? I can't remember if there are other wiki docs on translating.

craigh:
so three places I think:
1. https://github.com/zikula/core/blob/1.4/src/docs/en/dev/UPGRADE-1.4.0.md
2. https://github.com/zikula/core/wiki/Translating
3. changelog

craigh:
ping @shefik - can you take care of these other items so we can merge this?

shefik:
This wiki has been updated:

```
https://github.com/zikula/core/wiki/Translating
```

shefik:
UPGRADE-1.4.0.md has been updated.

Guite:
Thank you @shefik

PR #2433, opened 2015-05-04
Form should use "help-block" instead.

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Resources/views/Install/dbcreds.html.twig
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Resources/views/Install/dbcreds.html.twig
@@ -51,7 +51,7 @@
             <div class="col-sm-9">
                 {{ form_errors(form.database_name) }}
                 {{ form_widget(form.database_name) }}
-                <p class="help-block">{{ __('Ensure the database is in UTF-8 format.') }}</p>
+                <p class="help-block"><span class="text-danger">{{ __('Ensure the database is in UTF-8 format.') }}</span></p>
             </div>
         </div>
         <div class="form-group">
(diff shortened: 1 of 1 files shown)


src/lib/Zikula/Bundle/CoreInstallerBundle/Resources/views/Install/dbcreds.html.twig
```diff
@@ -51,9 +51,7 @@
             <div class="col-sm-9">
                 {{ form_errors(form.database_name) }}
                 {{ form_widget(form.database_name) }}
-                <em class="text-danger">
-                    {{ __('Ensure the database is in UTF8 format.') }}
-                </em>
+                <p class="help-block">{{ __('Ensure the database is in UTF-8 format.') }}</p>
```
craigh: help block will not color the text red like I wished

shefik: Does it really need to be red? It's "help" text, so shouldn't there consistent treatment with other forms?

craigh: I'm allowed to want it to be red :wink: I want it red because it is crucial.

shefik: Well, the red color is not the only thing. It's also styling it as italic. The help block is not italic by default.

shefik: @craigh I updated the pull request. Both classes are used now, so your desired color is applied.

PR #2434, opened 2015-05-04
Adding support for umask

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/app/config/parameters.yml
+++ b/src/app/config/parameters.yml
@@ -27,3 +27,14 @@ parameters:
     url_secret:        ~
 
     zikula.session.name: _zsid
+    
+    # Changing the current umask to the octal "022" will set directory permissions
+    # to "755", and file permissions to "644" (for directories such as
+    # "app/cache" and "app/logs"). Changing the current umask to the octal "02"
+    # will set directory permissions to "775", and file permissions to "664".
+    # Changing the current umask to the octal "0" will set (insecure) directory
+    # permissions to "777", and file permissions to "666". Note that using the ACL
+    # is recommended when you have access to them on your server, because changing
+    # the umask is not thread-safe. Set this value to "~" (NULL), to disable support
+    # for umask.
+    umask:             ~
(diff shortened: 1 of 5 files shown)


cmfcmf:
I don't think a new setting should be introduced in a legacy config file.

shefik:
@cmfcmf I tested other areas where this particular setting can be placed. However, it only seems to work immediately after the ZikulaKernel. Do you have a suggestion where to place it instead? The particular setting is needed, because the cache files are being created with insufficient permissions. For example, in my server environment, I need the cache files to have "group" permissions, and not just "user" permissions.

craigh:
> I don't think a new setting should be introduced in a legacy config file.

I agree this should not be a legacy feature. 

> Do you have a suggestion where to place it instead? 

probably [the config](https://github.com/zikula/core/blob/1.4/src/app/config/config.yml)

shefik:
@craigh @cmfcmf Updates have been made, to use YAML instead.

craigh:
@shefik have you tested this with a fresh install and upgrade from 13x?

shefik:
@craigh I have not tested with an upgrade, but it does work with a clean install of the latest build.

craigh:
could you please test the upgrade?

craigh:
> could you please test the upgrade?

ping @shefik

craigh:
@shefik. please test the upgrade.

craigh:
this requires a rebase.

craigh:
ping @shefik please rebase this so it can be merged.

craigh:
done in https://github.com/zikula/core/commit/dfb7017aedbe3e3753c09d14999f7081dc44ca08

PR #2439, opened 2015-05-08
Switching from text field to select, for the options of cache lifetime.

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/legacy/viewplugins/function.html_select_cache_lifetime.php
+++ b/src/lib/legacy/viewplugins/function.html_select_cache_lifetime.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Copyright Zikula Foundation 2009 - Zikula Application Framework
+ *
+ * This work is contributed to the Zikula Foundation under one or more
+ * Contributor Agreements and licensed to You under the following license:
+ *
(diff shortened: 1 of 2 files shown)


cmfcmf:
This could potentially break custom cache times like 60 seconds, etc. I vote for postponing this to 1.4.1. I generally don't want to see too many more features in 1.4.0, because we basically already are in the release process and have kind of a feature freeze.

shefik:
@cmfcmf I agree with your assessment. But, my attempt was to make it easier for site admins to change cache lifetime, without having to do calculations to determine what the values should be. It's easier to just have a drop-down menu. I fine with postponing to 1.4.1, as long as we re-visit this discussion.

shefik:
Perhaps the select menu can be transformed to have the pre-populated values, but also allow a custom value to be entered.

craigh:
> Perhaps the select menu can be transformed to have the pre-populated values, but also allow a custom value to be entered.

@shefik can you make these adjustments please?

PR #2447, opened 2015-05-10
Forward compatibility updates

craigh:
This PR fixes and adds several items that were encountered when creating a new "2.0.0-style" module. Several items are addressed that have not yet been required in "1.4.0-type" modules.

 - twig "plugins" (tags & filters)
 - new AbstractBlockController
 - CategoryType symfony FormType
 - automatic form theming for zikula admin forms

There are still some outstanding issues

 - ~~changelog~~
 - ~~CategoryType does not set selected values if multiple=true~~
 - does not work with multiple registries
 - ~~not all pager templates are migrated yet~~

There *should not* be any ill-effects that might change the existing codebase since much of this is new code that is not in use unless Twig or Symfony Forms is used. **New 'eyes' on the code is appreciated however before this is integrated.**

I will push all my changes to ZikulaPagesModule to demonstrate usage of these items immediately.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | n/a
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | yes :smile:


CHANGELOG-1.4.md
```diff
@@ -117,8 +117,11 @@ Features:
...
   - Auto-creates `custom_parameters.yml` and `personal_config.php`
 - Include Zikula/Wizard - a library to assist in multi-stage user interaction (used in installer)
+- Add Zikula/Component/SortableColumns - helps manage sortable column headers in datatables.
 - Added multilingual support for site name, site description and site meta tags (#2316).
 - Added view plugin {langchange} for switching language, function with shorturls enabled (#2364)
 - Added view plugin {moduleheader} to unify module headers and make styling at one place - moduleheader.tpl (#2372).
 - Added support Webshim, which is a polyfill library that enables you to reliably use HTML5 features across browsers, even if native support is lacking. (#2377)
 - Added mmenu js library and smarty template plugin to create a hidden admin panel
+- Added several Twig tags and filters to duplicate some legacy functionality
+- Added automatic form-theming (bootstrap-style) for Twig-based admin forms
```
cmfcmf: Awesome, an actual changelog update! :laughing:

src/system/Zikula/Module/SearchModule/AbstractSearchable.php
```diff
@@ -13,6 +13,7 @@
 
 namespace Zikula\Module\SearchModule;
 
+use Symfony\Component\DependencyInjection\ContainerInterface;
```
cmfcmf: Now that I see the new Search implementation again, I think it would be better to change it to behave like described here: http://symfony.com/doc/current/components/dependency_injection/tags.html

Then, no abstract classes are needed, just a single search interface for all bundles and modules. (will add a longer description ASAP).

src/system/Zikula/Module/CategoriesModule/Form/DataTransformer/CategoryModelTransformer.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param null $array
+     * @return \Zikula\Module\CategoriesModule\Entity\CategoryEntity
+     */
+    public function transform($array)
+    {
+        return !empty($array) ?
+            isset($array[$this->registryId]) ?
+                $array[$this->registryId]
+                : array()
+            : array();
```
cmfcmf: One or two lines / statements more would make this more readable :boat:

src/system/Zikula/Module/CategoriesModule/Form/Type/CategoryType.php
```diff
@@ -0,0 +1,122 @@
...
+            'query_builder' => function (EntityRepository $er) {
+                return $er->createQueryBuilder('c')
+                    ->where('c.parent = :parent')
+                    ->setParameter('parent', $this->parentId)
+                    ->orderBy('c.name', 'ASC');
+            },
+            'property' => "display_name[$this->lang]",
+            'multiple' => false,
+            'required' => false,
+            'placeholder' => __('Choose a category'),
```
cmfcmf: Missing domain?

src/lib/legacy/viewplugins/function.assignedcategorieslist.php
```diff
@@ -59,11 +59,19 @@ function smarty_function_assignedcategorieslist($params, Zikula_View $view)
...
             foreach ($params['categories'] as $category) {
-                if (!is_object($category) || !is_object($category->getCategory())) {
+                if (!is_object($category)) {
                     continue;
                 }
-                $name = $category->getCategory()->getName();
-                $display_name = $category->getCategory()->getDisplayName();
+                if ($category instanceof \Zikula\Core\Doctrine\Entity\AbstractEntityCategory) {
+                    if (!is_object($category->getCategory())) {
+                        continue(2);
```
cmfcmf: Why is there a 2? I only see one loop to continue.

src/lib/Zikula/Core/Theme/Asset.php
```diff
@@ -83,8 +83,10 @@ public function resolvePath($path)
     private function getAssetPath($bundleName, $assetPath)
     {
         $bundle = $this->kernel->getBundle($bundleName);
-        $assetDir = 'bundles/';
-        $targetDir = $assetDir.preg_replace('/bundle$/', '', strtolower($bundle->getName()));
+        $bundleType = method_exists($bundle, 'getNameType') ? strtolower($bundle->getNameType()) : 'bundle';
+
+        $assetDir = "{$this->webDir}/{$bundleType}s/";
+        $targetDir = $assetDir . preg_replace("/{$bundleType}$/", '', strtolower($bundle->getName()));
```
cmfcmf: `str_replace` would be more simple and faster(?).

src/lib/Zikula/Core/Controller/AbstractBlockController.php
```diff
@@ -0,0 +1,74 @@
...
+     * @param array  $args   Arguments.
+     *
+     * @throws \BadMethodCallException If no event responds.
+     *
+     * @return string Data.
+     */
+    public function __call($method, $args)
+    {
+        $event = new \Zikula\Core\Event\GenericEvent($this, array('method' => $method, 'args' => $args));
+        $this->get('eventmanager')->dispatch('block.method_not_found', $event);
```
cmfcmf: Is this a new or old event?

src/lib/Zikula/Component/SortableColumns/SortableColumns.php
```diff
@@ -0,0 +1,201 @@
...
+ * Please see the NOTICE file distributed with this source code for further
+ * information regarding copyright and licensing.
+ */
+
+namespace Zikula\Component\SortableColumns;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Symfony\Component\Routing\RouterInterface;
+
+class SortableColumns {
```
cmfcmf: `{` should be on the next line in most of your new classes.

src/system/Zikula/Module/CategoriesModule/Form/Type/CategoryType.php
cmfcmf: Ahh wait, it's a system module, so it doesn't need it.

src/lib/Zikula/Core/Theme/Asset.php
craigh: just adjusting code from drak here. could change though, I guess

src/lib/Zikula/Core/Controller/AbstractBlockController.php
craigh: new event (see namespace).

src/lib/Zikula/Component/SortableColumns/SortableColumns.php
craigh: PhpStorm auto-formatting :laughing:

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/PagerExtension.php
```diff
@@ -0,0 +1,399 @@
...
+                if (!empty($arg)) {
+                    $argument = explode('=', $arg);
+                    $pager['args'][$argument[0]] = $argument[1];
+                }
+            }
+        }
+
+        //also $_POST vars have to be considered, i.e. for search results
+        $allVars = ($params['includePostVars']) ? array_merge($_POST, $_GET, $routeParams) : array_merge($_GET, $routeParams);
+        foreach ($allVars as $k => $v) {
```
cmfcmf: This can probably vanish.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/PagerExtension.php
```diff
@@ -0,0 +1,399 @@
+<?php
```
cmfcmf: I would drop support for not-routed pages.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/PagerExtension.php
```diff
@@ -0,0 +1,399 @@
...
+     *
+     * @param array $params All attributes passed to this function from the template.
+     * @return string
+     */
+    public function pager($params)
+    {
+        /** @var Request $request */
+        $request = $this->container->get('request');
+        if (!isset($params['rowcount'])) {
+            trigger_error(__f('Error! in %1$s: the %2$s parameter must be specified.', array('pager', 'rowcount')));
```
cmfcmf: Is trigger_error the right thing to do?

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/PagerExtension.php
cmfcmf: I think the pager should be re-done / have more comments, etc. I remember problems with 0 vs. 1 as starting position.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/HookExtension.php
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        // create event and notify
+        $hook = new DisplayHook($id, $urlObject);
+        $this->container->get('hook_dispatcher')->dispatch($eventName, $hook);
+        $responses = $hook->getResponses();
+
+        $output = '';
+        foreach ($responses as $result) {
+            $output .= "<div class=\"z-displayhook\">$result</div>\n";
```
cmfcmf: Probably worth to check whether `$result` is empty and then skip appending to the output.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
```diff
@@ -192,12 +211,98 @@ public function arrayUnset($array, $key)
...
+
+    /**
+     * @todo this maybe should be rewritten to a Tag (Token Parer and Node Visitor) and called {% pageSetVar name = "value" %}
      *
-     * @return string The extension name
+     * @param null $name
+     * @param null $value
      */
-    public function getName()
+    public function pageSetVar($name = null, $value = null)
```
cmfcmf: Why do you use null? Aren't both mandatory?

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
```diff
@@ -192,12 +211,98 @@ public function arrayUnset($array, $key)
...
     {
-        return 'zikulacore';
+        if (empty($name) || empty($value)) {
+            throw new \InvalidArgumentException(__('Empty argument at') . ':' . __FILE__ . '::' . __LINE__);
+        }
+
+        if (in_array($name, array('stylesheet', 'javascript'))) {
+            $value = explode(',', $value);
```
cmfcmf: I vote for removing this. There is no harm in calling pageSetVar multiple times for multiple js / css files.

src/lib/Zikula/Core/Controller/AbstractBlockController.php
cmfcmf: Then it should probably be prepended by `zikula` or something?

src/system/Zikula/Module/SearchModule/AbstractSearchable.php
cmfcmf: The SearchModule would define a service called `zikulasearchmodule.searchable_chain` or something. This service will hold the list of all the searchable modules. To do so, every module who wants to use the new search stuff would need to create a service themself, with a tag like `<tag name="zikula.searchable" />` in it's service definition. The modules' services would need to implement a SearchableInterface which defines the needed methods (like `getResults`).
Then, in a compiler pass of the SearchModule, you would do what's done here http://symfony.com/doc/current/components/dependency_injection/tags.html#create-a-compilerpass.

Then, when a search is issued, the SearchModule can use it's own service, which now has a list of all the searchable services, and get the results one by one.

The advantage is that literally every bundle (not only modules) can provide a search. Also, the bundles don't need to extend some abstract class.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/PagerExtension.php
craigh: the whole method needs to be updated to eliminate BC, but didn't take to do so yet. (see `@todo` in header)

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/PagerExtension.php
craigh: temporarily, yes :stuck_out_tongue_closed_eyes:

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
craigh: copying from legacy function, mainly. would like to rewrite this to a Tag and not a function anyway.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
craigh: not sure if I agree. @Guite ? @shefik ? other opinions?

src/system/Zikula/Module/SearchModule/AbstractSearchable.php
craigh: An advantage of the current approach would be that methods in the abstract are available.

I'm not opposed to your idea, but I am unwilling to code it. If you are willing to do so, please do so.

craigh: it would mean refactoring current implementations in multiple modules as well as /system modules. I think Pages, Dizkus and maybe a few others (News??) also have the current implementation.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
cmfcmf: Remember we had that problem where a `,` was part of the url?

src/lib/legacy/viewplugins/function.assignedcategorieslist.php
craigh: can't remember. I think the `if` are also continued?

src/lib/Zikula/Core/Controller/AbstractBlockController.php
craigh: I misunderstood your question. I believe the event is a 1.3. name.

src/lib/legacy/viewplugins/function.assignedcategorieslist.php
cmfcmf: I don't think so, but I'm not 100% sure.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/PagerExtension.php
craigh: yes. see other note.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
shefik: @cmfcmf You are right about the comma. It's because of the Google fonts URL.

craigh: can't remember :forgetful:

craigh: but was never changed in smarty plugin...

cmfcmf: it was: https://github.com/zikula/core/blob/1.4/src/lib/legacy/viewplugins/function.pageaddvar.php#L25

shefik: For the multiple addVar calls, what happens? Is there a performance difference?

craigh: ok, so I was wrong :facepunch: 
suggestions on a proper solution then? just eliminate the army as a possibility?

src/lib/Zikula/Core/Controller/AbstractBlockController.php
craigh: The event name [is a 1.3 (legacy) name](https://github.com/zikula/core/blob/1.4/src/lib/legacy/Zikula/Controller/AbstractBlock.php#L81)

how about removing the magic method altogether?

cmfcmf: That's what I meant with "new event" :laughing:

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
craigh: By the way - this comment thread is based of the `pageSetVar()` function but you are relating it to the legacy `{pageaddvar}` template plugin. There is a companion `pageAddVar()` method here as well. It also has the `$raw` optional parameter.

I'm not saying the action being discussed (remove array option) is wrong, I'm just pointing out a possible misunderstanding.

shefik:
@craigh Are there any database changes in this particular code update?

craigh:
no

Issue #2448, opened 2015-05-11
Override a Controller Class


shefik:
Previously, I was able to use the event `module_dispatch.custom_classname` to override a Controller class. Now, when using a module converted to Symfony routes, `module_dispatch.custom_classname` does not appear seem to be called anymore for the Controller. It is only being called for the Api (for the `getLinks` function).

Which event can be used instead?

craigh:
check out

http://symfony.com/doc/current/cookbook/bundles/override.html
http://symfony.com/doc/current/cookbook/bundles/inheritance.html

shefik:
@craigh Zikula modules uses non-service controllers, right? So, these docs won't be applicable:

http://symfony.com/doc/current/cookbook/bundles/override.html#services-configuration

shefik:
Per the docs here:

http://symfony.com/doc/current/cookbook/bundles/inheritance.html

The sample code says `Start by registering the FOSUserBundle as the "parent" of your bundle`

```
public function getParent()
    {
        return 'FOSUserBundle';
    }
```

So, I tested by by adding the function `getParent()` to a module's controller class. However, `getParent()` doesn't get called by the module. So, some other method is needed.

shefik:
Looks like the `getParent()` needs to be set in this file to get called - `/lib/Zikula/Core/AbstractBundle.php`.

craigh:
I'm removing the BCBreak from this because old methods still work as they did before. it is only when using new technologies are there problems, so IMO this is not a BC Break.

craigh:
closing. If an extension author wishes, they can use the `setParent()` method in their bundle class.

PR #2449, opened 2015-05-12
Adding support for getting the parent class of a bundle

shefik:
Based on the Symfony docs referenced here:

http://symfony.com/doc/current/cookbook/bundles/inheritance.html

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | https://github.com/zikula/core/issues/2448
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Core/AbstractBundle.php
+++ b/src/lib/Zikula/Core/AbstractBundle.php
@@ -6,6 +6,7 @@ use Symfony\Component\Console\Application;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
+use Symfony\Component\Yaml\Yaml;
 
 abstract class AbstractBundle extends Bundle
 {
@@ -17,6 +18,15 @@ abstract class AbstractBundle extends Bundle
     protected $booted = false;
 
     private $basePath;
+
+    public function getParent()
+    {  
+        if (is_readable($file = $this->getConfigPath().'/bundle.yml')) {
+            $bundleConfig = Yaml::parse(file_get_contents($file));
+            
+            return $bundleConfig[strtolower($this->name)]['parent'];
+        }
+    }
 
     public function isBooted()
     {


shefik:
A listener (such as `module_dispatch.custom_classname`) can't be used anymore for this type of class override. Instead, use inheritance to override, by creating a new YAML file at this example module path:

`modules/Example/Resources/config/bundle.yml`

The file contents should be based on this sample:

```
zikulaexamplemodule:
    parent: ZikulaParentModule
```

Then, in your controller, instead of extending `\Zikula_AbstractController`, extend a class within that parent module. The parent module's class should be extend `\Zikula_AbstractController` as normal.

shefik:
The code I committed works for one module to directly extend another module. However, my need is to have one module directly extend multiple modules. Apparently, this is not supported in Symfony2, because only one bundle can be extended.

http://stackoverflow.com/questions/9109669/how-can-i-extend-a-bundle-more-than-once
http://stackoverflow.com/questions/15428516/symfony2-inheriting-more-than-one-bundle

shefik:
I was able to accomplish what I wanted to do, by simply having one module extend multiple modules as-is. It wasn't necessary to set a parent in Symfony to achieve this. Using this sample code in the child:

```
public function exampleAction(Request $request)
    {
        $response = parent::exampleAction($request);

        // ... do custom stuff
        return $response;
    }
```

However, the code in this pull request is still useful, if anyone needs to extend one module with another module. That functionality was missing in the Zikula core.

craigh:
@shefik - why do we need a 'dynamic' `getParent` method? Why not simply ask the bundle(module) author to declare their own method with the parent? Maybe I'm not understanding the value here.

craigh:
ping @shefik

shefik:
@craigh I am not using the code in this pull request anymore in my own use case, due to the issue I mentioned earlier:

`The code I committed works for one module to directly extend another module. However, my need is to have one module directly extend multiple modules. Apparently, this is not supported in Symfony2, because only one bundle can be extended.`

I am now using a different solution to achieve what I want, which is also described above (pull request not needed).

This pull request can be closed, if you don't think there is any need for this particular code.

PR #2469, opened 2015-05-25
CLI problems with mod_suexec

rallek:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | [no]

Added a Comment for the case of mod_suexec

### DIFF
--- a/INSTALL-1.4.0.md
+++ b/INSTALL-1.4.0.md
@@ -111,6 +111,8 @@ $ php app/console zikula:install:finish
 ```
 Open your browser and login!
 
+Attention: If you have ``mod_suexec`` installed at Apache the CLI will run into permission problems. If you are not shure see phpinfo. mod_suexec often is used at shared hosting. In this case you have to use the Web Installer.
+
 
 <a name="final"></a>
 5. Final Note
--- a/INSTALL-1.4.0.md
+++ b/INSTALL-1.4.0.md
@@ -111,7 +111,7 @@ $ php app/console zikula:install:finish
 ```
 Open your browser and login!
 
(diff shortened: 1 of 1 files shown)


INSTALL-1.4.0.md
```diff
@@ -111,6 +111,8 @@ $ php app/console zikula:install:finish
 ```
 Open your browser and login!
 
+Attention: If you have ``mod_suexec`` installed at Apache the CLI will run into permission problems. If you are not shure see phpinfo. mod_suexec often is used at shared hosting. In this case you have to use the Web Installer.
+
```
craigh: "sure" not "shure"

*have to* use the web installer? no choice? @Guite @cmfcmf

rallek: the CLI will create folders with the wrong ownership and wrong permissions.

cmfcmf:
I am not sure if this really is the problem of `mod_suexec`.

craigh:
> I think that's always the case when running from CLI and the cli user differs from the webserver's user.

hmm. **that** could be a problem, no? I wonder if that can be corrected?

rallek:
mod_suexec was the idea of @shefik.

cmfcmf:
See http://symfony.com/doc/current/book/installation.html#checking-symfony-application-configuration-and-setup. Symfony has the exact same problem. They also provide a CLI installer.

craigh:
@rallek please add this link to the documentation also.

craigh:
I guess the "good" news here is that Drupal (and other highly used CMS products that are migrating to or currently are based on Symfony) will all have the same problem :confused:

craigh:
merging this but plan to reformat the link.

PR #2475, opened 2015-05-26
Update bootstrap.tpl

rallek:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | [no]
Adding the hamburger symbol for mobile view

### DIFF
--- a/src/system/Zikula/Module/BlocksModule/Resources/views/Block/Menutree/bootstrap.tpl
+++ b/src/system/Zikula/Module/BlocksModule/Resources/views/Block/Menutree/bootstrap.tpl
@@ -1,5 +1,13 @@
 <nav class="navbar navbar-default" role="navigation">
     <div class="container">
+        <div class="navbar-header">
+          <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-collapse">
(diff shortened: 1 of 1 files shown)


Kaik:
Hmm I do not know if this is good idea to add 'hamburger' menu by default. Not all menu's should be collapse able. We would need to add setting in block to turn it off/on.

rallek:
@Kaik You are right. Turning on/off would be great. But unfortunally  I do not have enough skills. Is it a solution just to add a second hamburger-bootstrap.tpl?

Kaik:
>Is it a solution just to add a second hamburger-bootstrap.tpl

As long as there is a way to chose this template in block modify it will be good temporary solution :)

rallek:
I will redo these things. I tested localy the way to have a second template. While doing this I found some other smal issues.
I will do a new PR with two layouts and some small improvements later. First I have to bring the fork to my computer :-)

Issue #2452, opened 2015-05-13
Using the "moduleheader" plugin when render cache is enabled


shefik:
Using the `moduleheader` plugin is not compatible when a module's render cache is enabled.

Example code:

```
{module header modname=$module type=$type title=$templatetitle setpagetitle=true insertstatusmsg=true}
```

The output of the plugin is being cached, which means the module links are being cached (and not respecting permissions). Also, the status message is being cached also, and being persistently displayed.

craigh:
can you look into a solution? caching is voodoo to me.

Guite:
Postponed to 1.4.1

craigh:
marking this as won't fix. Solution/workaround is to use new Twig solutions.

craigh:
refs #2793

PR #2479, opened 2015-05-29
Module Header

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | https://github.com/zikula/core/issues/2452
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/legacy/viewplugins/function.moduleheader.php
+++ b/src/lib/legacy/viewplugins/function.moduleheader.php
@@ -59,7 +59,7 @@ function smarty_function_moduleheader($params, Zikula_View $view)
     }
     $titlelink = isset($params['titlelink']) ? $params['titlelink'] : false;
 
-    $renderer = Zikula_View::getInstance('Theme');
+    $renderer = Zikula_View::getInstance('ZikulaThemeModule');
     $renderer->setCaching(Zikula_View::CACHE_DISABLED);
 
     $renderer->assign('userthemename', UserUtil::getTheme());
--- a/src/lib/legacy/viewplugins/function.moduleheader.php
+++ b/src/lib/legacy/viewplugins/function.moduleheader.php
@@ -29,7 +29,7 @@
  *
  * @return string A formatted string containing navigation for the module admin panel.
  */
-function smarty_function_moduleheader($params, $view)
+function smarty_function_moduleheader($params, Zikula_View $view)
 {
     if (!isset($params['modname']) || !ModUtil::available($params['modname'])) {
         $params['modname'] = ModUtil::getName();
--- a/src/lib/legacy/viewplugins/function.moduleheader.php
+++ b/src/lib/legacy/viewplugins/function.moduleheader.php
@@ -43,7 +43,7 @@ function smarty_function_moduleheader($params, $view)
     $putimage = isset($params['putimage']) ? $params['putimage'] : false;
     $setpagetitle = isset($params['setpagetitle']) ? $params['setpagetitle'] : false;
     $insertstatusmsg = isset($params['insertstatusmsg']) ? $params['insertstatusmsg'] : false;
-    $cutlenght = isset($params['cutlenght']) ? $params['cutlenght'] : 20;
+    $cutlength = isset($params['cutlength']) ? $params['cutlength'] : 20;
     if ($putimage) {
         $image = isset($params['image']) ? $params['image'] : ModUtil::getModuleImagePath($params['modname']);
     } else {
@@ -67,8 +67,8 @@ function smarty_function_moduleheader($params, $view)
     $renderer->assign('type', $params['type']);
     $renderer->assign('title', $params['title']);
     $renderer->assign('titlelink', $titlelink);
-    $renderer->assign('truncated', mb_strlen($params['title']) > $cutlenght);
-    $renderer->assign('titletruncated', mb_substr($params['title'], 0, $cutlenght) . '...');
+    $renderer->assign('truncated', mb_strlen($params['title']) > $cutlength);
+    $renderer->assign('titletruncated', mb_substr($params['title'], 0, $cutlength) . '...');
     $renderer->assign('setpagetitle', $setpagetitle);
     $renderer->assign('insertstatusmsg', $insertstatusmsg);
     $renderer->assign('menufirst', $menufirst);


shefik:
Note: This pull request just fixes a typo and set a type for $view, but it does not fix the issue described in https://github.com/zikula/core/issues/2452. After further testing, it looks like caching is only respected on first level templates, not second level templates (which is what the `module header` plugin is using.

For example this can be considered a first-level template:

A module direct calls to the template `User/menu.tpl`, containing the code for `{insert name='getstatusmsg'}`. When render cache is enabled, the status message does not get cached.

However, the `moduleheader` plugin is placed in the template `User/menu.tpl` instead via the code such as `{moduleheader modname=$module type=$type title=$templatetitle setpagetitle=true insertstatusmsg=true}`. When the plugin code is executed, it calls to another template within the Theme module at `system/Zikula/Module/ThemeModule/Resources/views/moduleheader.tpl`. Even though the `moduleheader` plugin has code within it that explicitly says disable cache, it is still being cached on the subsequent call to `system/Zikula/Module/ThemeModule/Resources/views/moduleheader.tpl`.

shefik:
I tried re-writing the plugin to use the existing $view instance, instead of getting a new instance. For example:

```
    $view->setCaching(Zikula_View::CACHE_DISABLED);

    $view->assign('userthemename', UserUtil::getTheme());
    $view->assign('modname', $params['modname']);
    $view->assign('type', $params['type']);
    $view->assign('title', $params['title']);
    $view->assign('titlelink', $titlelink);
    $view->assign('truncated', mb_strlen($params['title']) > $cutlength);
    $view->assign('titletruncated', mb_substr($params['title'], 0, $cutlength) . '...');
    $view->assign('setpagetitle', $setpagetitle);
    $view->assign('insertstatusmsg', $insertstatusmsg);
    $view->assign('menufirst', $menufirst);
    $view->assign('image', $image);
    
    $template = 'file:'.$kernel->getRootDir().'/../'.ModUtil::getModuleRelativePath('ZikulaThemeModule').'/Resources/views/moduleheader.tpl';
 
    if ($assign) {
        $view->assign($assign, $view->fetch($template));
    } else {
        return $view->fetch($template);
    }
```

Although this does fix the caching issue, as well as the multi-level template calls, it introduces a new problem where `{nocache}` code might be output on the page, if a module template has `{pageaddvar}` calls.

shefik:
Based on the way caching currently works in Zikula, I don't think the `moduleheader` plugin can be used at all, because fixing it will cause other parts of the core to break. Unless anyone else has another solution, I suggest removing it from the core.

craigh:
it cannot be removed for at least two reasons:

1) maintain BC
2) every admin page uses it as part of the adminheader

shefik:
But, wasn't this a new plugin that got recently added?

shefik:
It was just added in March 2015 - https://github.com/zikula/core/pull/2372. It hasn't been included in any officially release yet, so there wouldn't be a BC break.

shefik:
This is the code that appears to work well, when both render cache and theme cache are both enabled:

```
{insert name='getstatusmsg'}
<div class="{if ($type == 'admin')}z-admin-content-modtitle{else}z-modtitle{/if}">
    <h2><img src="modules/{$modinfo.directory}/Resources/public/images/admin.png" alt="{$modinfo.displayname}" class="middle" /> {$modinfo.displayname}</h2>
</div>
{nocache}{modulelinks modname=$modinfo.name type=$type}{/nocache}
```

Insert the above code directly in the templates `User/menu.tpl` and `Admin/menu.tpl`.

craigh:
> It was just added in March 2015

You are correct. I was mistaken. There is no BC concern. But it has been implemented thought the core.

Are you having trouble with caching in admin templates at all? My understanding is the admin caching is disabled.

shefik:
That is true, admin pages have caching disabled by default, so there is no issue there. However, I came across the problem, when I was attempting to be consistent on the user pages, and re-use the same menu code. The plugin can't be used on non-admin pages properly. And, I don't think it's fixable due to the way cashing works in Zikula at the moment. It's just not compatible to use.

craigh:
I think we need @nmpetkov input on this please. since he is the one that did it.

one question would be, did caching work before this plugin was implemented? you gave an example of working code above, can this be implemented into the plugin template?

shefik:
Instead of using this plugin, the template code posted here in this particular comment works across all admin and user pages, even when render cache is enabled:

https://github.com/zikula/core/pull/2479#issuecomment-106669401

Also, if this pull request is accepted, then the same code will work when theme cache is enabled simultaneously:

https://github.com/zikula/core/pull/2480

shefik:
The example code I posted can't be used with the plugin template, because of the way Zikula processes `insert` and `nocache`. It needs to be placed within the module template. It's also an issue, where the plugin has to create a new instance of the render view.

nmpetkov:
I'll test again the plugin with caching enabled. As I remember caching in this case is turned off (it is possible to turn of caching in a view instance  - there is a parameter for this purpose).

Main reason to create this plugin, was to have only one place, where to make design changes for the module headers (in one template).

I'm using it several months without any problems (Zikula 1.3.10). I'm using only theme caching, and for main (home) page only, as all my attempts to use caching with Zikula 1.3.9 (10) are with problems.

shefik:
@nmpetkov Yes, caching is turned off in the plugin. However, it is not effective, due to the issues I described here:

https://github.com/zikula/core/pull/2479#issuecomment-106645973

The code and pull request I suggested works for me under the latest build of Zikula 1.4.0, with render cache and theme cache both enabled.

craigh:
`<h2>{$modinfo.displayname} {$modinfo.displayname}</h2>`

@shefik this won't work for `/system` modules.

craigh:
> I'll test again the plugin with caching enabled.

ping @nmpetkov

craigh:
I agree that there is typo in the code, but it is part of the API. "fixing" the typo is technically a BC break. Since all the smarty plugins are technically deprecated anyway, I don't think it is worth merging a fix.

craigh:
@nmpetkov @shefik it sounds like you both disagree on the validity of the bug and fix. please test and communicate with each other to resolve the issue. Then we can merge a fix if required.

PR #2481, opened 2015-05-29
Adding Proper Support for Fonts

shefik:
Adding support for page var "font", and correcting error, where @import was not supporting remote URLs.

Example use in a template:

`{pageaddvar name='font' value='//fonts.googleapis.com/css?family=Oswald:300,400,700'}`

Any remote font URL can be used. It does not need to be a Google font. All fonts will be loaded before other stylesheets, since it is required for fonts to always load first for proper styling.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/util/JCSSUtil.php
+++ b/src/lib/util/JCSSUtil.php
@@ -103,10 +103,23 @@ class JCSSUtil
         // update stylesheets as there might be some additions for js
         $stylesheets = array_merge((array)$stylesheets, (array)PageUtil::getVar('stylesheet'));
         $stylesheets = self::prepareStylesheets($stylesheets, $themeinfo, $isAdminController);
+        
         if ($combine) {
             $javascripts = (array)self::save($javascripts, 'js', $cache_dir);
             $stylesheets = (array)self::save($stylesheets, 'css', $cache_dir);
         }
+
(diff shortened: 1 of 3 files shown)


craigh:
what was the old solution? why doesn't it work anymore?

craigh:
> what was the old solution? why doesn't it work anymore?

ping @shefik response?

src/lib/util/JCSSUtil.php
```diff
@@ -675,19 +688,24 @@ private static function readfile(&$contents, $file, $ext)
...
-                                    // fix url
-                                    $url = dirname($file) . '/' . $url;
-                                    if (!$wasCommentHack) {
-                                        // clear buffer
-                                        $contents[] = $newLine;
-                                        $newLine = "";
-                                        // process include
-                                        self::readfile($contents, $url, $ext);
+                                    
+                                    if (!is_readable($url)) {
```
Guite: Does this check for whether it is a remote url? Can this be done in a cheaper way avoiding file system access?

Issue #2484, opened 2015-05-31
Update to Symfony 2.7 LTS


Guite:
1.4 should support Symfony 2.7 because it is a LTS release.

craigh:
symfony 2.7 was [released yesterday 30 May 2015](http://symfony.com/blog/symfony-2-7-0-released)

craigh:
[upgrade docs](https://github.com/symfony/symfony/blob/2.7/UPGRADE-2.7.md)

craigh:
closed by #2486

Issue #2483, opened 2015-05-30
composer no longer installing zikula-modules in /modules


craigh:
composer no longer installing zikula-modules (legal/profile/breakit) correctly at least not on `composer update`

Composer version 1.0-dev (e64470c987fdd6bff03b85eed823eb4b865a4152) 2015-05-28 14:52:12

it copies repos to `/vendors` and not `/modules`

Guite:
It still works in Jenkins using a fresh workspace, so `composer install` is not affected.

Closing this one.

craigh:
I'm not sure I agree here. I just downloaded from CI. There is a full copy of both Profile and Legal in the `/vendor/zikula` and another full copy in `/modules/zikula`. This is not correct IMO. The full copy should only be in `/modules/zikula`. (also, there are no assets in `/web/modules`)

craigh:
and `composer update` is what the ticket is about.

cmfcmf:
Probably fixed by https://github.com/RobLoach/component-installer/pull/87/files

craigh:
> Probably fixed by...

if so, then some stuff must be removed from #2486 @Guite

Guite:
Yes, #2486 contains a hacky workaround at the moment. Will remove this after the PR mentioned above works.

craigh:
indeed https://github.com/RobLoach/component-installer/pull/87/files has solved this issue.
@Guite - please update #2486

Guite:
done

PR #2486, opened 2015-06-06
Updated to Symfony 2.7 LTS

Guite:
This updates the 1.4 branch for using Symfony 2.7 LTS version.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | yes
| Tests pass?       | yes
| Fixed tickets     | #2484
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | yes


src/lib/Zikula/Core/AbstractBundle.php
```diff
@@ -106,9 +106,16 @@ public function getConfigPath()
...
-        return substr($path, strpos($path, $matches[1]), strlen($path));
+        $relativePath = substr($path, strpos($path, $matches[1]), strlen($path));
+
+        // hacky patch for modules installed by composer
+        if ($relativePath == 'vendor') {
+            $relativePath = 'modules';
+        }
+
+        return $relativePath;
     }
```
Guite: This relates to #2483

craigh:
did you check on all the new forms in the InstallerBundle and elsewhere (Pages, BreakIt, etc)

src/lib/Zikula/Bundle/CoreBundle/Templating/Asset/PathPackage.php
```diff
@@ -2,25 +2,25 @@
...
 
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\Templating\Asset\PathPackage as BasePathPackage;
+use Symfony\Component\Asset\PathPackage as BasePathPackage;
 
 /**
  * The path package adds a version and a base path to asset URLs.
+ *
+ * TODO This class can probably be removed.
+ * The setting "templating.asset.path_package.class" is already disabled in app/config/config.yml.
```
craigh: then why not remove it?

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/AssetsExtension.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Zikula\Bundle\CoreBundle\Twig\Extension;
+
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Zikula\Bundle\CoreBundle\Twig;
+
+use Symfony\Bundle\TwigBundle\Extension\AssetsExtension as Base;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class AssetsExtension extends Base {
```
craigh: I'm not a fan of the alias 'Base' because one cannot discern any meaning from the word. In this case something like `SymfonyAssetsExtension` is more descriptive.

craigh:
> did you check on all the new forms

I read the upgrade docs more closely and see that all the Forms changes are only deprecations and not complete removal, so I see other changes are not yet *required*

craigh:
this needs to "turn off" all the Deprecated warnings.

screen shot 2015-06-07 at 4 10 43 pm

It takes too long to load stuff.

craigh:
or refactor as needed.

craigh:
Silencing the warnings:
http://symfony.com/doc/current/cookbook/upgrade/deprecation_warnings.html

http://stackoverflow.com/questions/30572234/how-to-ignore-deprecated-messages-in-log-related-to-3rd-party-bundles

http://stackoverflow.com/questions/28850809/disable-deprecated-warning-in-symfony-2-7

src/app/config/config.yml
```diff
@@ -14,6 +14,10 @@ framework:
     validation:      { enabled: true, enable_annotations: true, api: auto }
     templating:
         engines: ['twig']
+        assets_base_urls: %assets_base_url%
```
Guite: There is a problem with the initial call of composer if this parameter has no value. If providing a value overriding it fails though (see bootstrap file below).

src/lib/bootstrap.php
```diff
@@ -19,6 +19,19 @@
...
 ZLoader::register($loader);
 
+// set current domain for using it as container parameter
+// http://symfony.com/doc/current/cookbook/configuration/external_parameters.html
+if (strpos(__DIR__, '/src/lib') !== false) {
+    $_SERVER['SYMFONY__ASSETS_BASE_URL'] = '//' . $_SERVER['HTTP_HOST'] . '/src/web';
+} else {
+    $_SERVER['SYMFONY__ASSETS_BASE_URL'] = '//' . $_SERVER['HTTP_HOST'] . '/web';
+}
+
```
Guite: In general this works fine. The problem is the initial usage of Composer, like mentioned above.
After providing a default value Symfony does not accept the dynamic setting anymore.

I also tried this:
```
$assetsBaseUrl = '//' . $_SERVER['HTTP_HOST'] . ((strpos(__DIR__, '/src/lib') !== false) ? '/src/web' : '/web');
$core->getContainer()->setParameter('assets_base_url', $assetsBaseUrl);
```

Guite:
I already added some error reporting changes to hide deprecation warnings on the actual site. Don't know if it can also be excluded from the developer toolbar.

src/lib/bootstrap.php
Guite: Added some information in Slack, too.

Guite:
Let's focus on getting the asset path thingy to work first so that this PR can be merged.
Afterwards we can look at the deprecation warnings within a separate ticket.

craigh:
just noticed you didn't use a topic branch for this...

src/lib/bootstrap.php
craigh: I have _**removed**_ all this and also the line in config.yml and everything seems to work... what am I missing?

craigh:
I just tried every option I could find for silencing the deprecation warnings and nothing worked.... :cry:

src/lib/bootstrap.php
Guite: I just did the following:
- clone my fork
- remove this code in bootstrap.php and the line in config.yml
- execute composer install
- call http://localhost/src/install

Then I see this:
screen_installer

Which was the very first reason for this topic ;-)

craigh: Try it with only removing the line from config.yml

Guite: When only the config line is removed then the code above has no purpose, because the server variable exported in bootstrap.php is never used.

Guite: Some background information:
https://github.com/symfony/symfony/issues/4883
https://github.com/symfony/symfony/pull/13234

craigh: > When only the config line is removed then the code above has no purpose

that parameter is loaded either way, it has default values. By adding the line you are simply changing the defaults. the lines in bootstrap should still work. did you try it?

Guite: `%assets_base_url%` is an own (a custom) parameter which has no default value. This is why composer fails when no value is set in the config. But as soon as I set a (default) value in parameters.yml the dynamic value from the bootstrap code is not set/used.

I will now check if it behaves differently if we remove the config parts and directly set the original value in bootstrap instead of introducing a custom parameter at all.
```
$core->getContainer()->setParameter('framework.templating.assets_base_urls', $assetsBaseUrl);
```

craigh: [assets_base_url is a framework config value](http://symfony.com/doc/current/reference/configuration/framework.html#assets-base-urls)

Guite: Seems I've found something... some config values have changed with 2.7... investigating.

Guite:
I think this can be merged now. It should work fine apart from the deprecation warnings (which we should approach independently).

craigh:
I would like to look at this again tonight before it is merged.

The deprecation warnings are a significant issue in the dev environment, causing significant slowdowns to the point of near unusability. Can you please look into this ASAP. I'm not sure we should separate that into an independent issue.

Guite:
On a new install I can not see any deprecated warnings in the toolbar at all. Which is also what I expect from https://github.com/zikula/core/pull/2486/files#diff-63a629f1abd9e33b945e10c22eb2f894 and https://github.com/zikula/core/pull/2486/files#diff-520f6cf34e6d282f21e8d5b55fdc2aad
I will remove these for test and see whether I can refactor a few of them without problems.

Guite:
On a new install I can not see any deprecated warnings in the toolbar at all.
Maybe they are related to 3rd party modules in your case?

Kaik:
Hi, I have removed all the files from src folder and composer.lock, reset branch and working tree hard (composer.json same as 1.4 branch) then composer install, paste config files and deprecated showed up ~1056

Modules are installed properly in /modules no additional copy in /vendor assets present in web/modules for all modules site is working

```
DEPRECATED - The Symfony\Component\Routing\Route::getPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the getPath() method instead.  +
DEPRECATED - The Symfony\Component\Routing\Route::setPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the setPath() method instead.  +
DEPRECATED - The Symfony\Component\Routing\Route::getPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the getPath() method instead.
DEPRECATED - The Symfony\Component\Routing\Route::setPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the setPath() method instead.
DEPRECATED - The Symfony\Component\Routing\Route::getPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the getPath() method instead.
DEPRECATED - The Symfony\Component\Routing\Route::setPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the setPath() method instead.
DEPRECATED - The Symfony\Component\Routing\Route::getPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the getPath() meth
```

@Guite @craigh

Guite:
We can not prevent all deprecation warnings. At the end they have a purpose also.

craigh:
sure, but this is an unreasonable number and the simple writing of the log causes the web browser to time out.

You said you weren't experiencing any. Are you in dev environment with debug on?

Guite:
Going to recheck and solve this later today.

craigh:
screen shot 2015-06-10 at 6 58 00 am

PR #2491, opened 2015-06-10
Deprecation error handling

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | #2486 
| License           | MIT
| Doc PR            | -
| Changelog updated | no


Guite:
Collecting related issues in vendor bundles here:

https://github.com/FriendsOfSymfony/FOSJsRoutingBundle/issues/196

Guite:
I think that some deprecated route information comes from the database.

Example content for the `methods` field:
```a:1:{i:0;s:3:"GET";}```

Example content for the `requirements` field:
```a:1:{s:7:"_method";s:3:"GET";}```

Whereby the second parameter is deprecated.

@cmfcmf How to handle that?

Guite:
@cmfcmf and I looked at this during the last hour. We did some minor refactorings in order to do what is possible.
One thing which is outside of our area is the `_method` depreciation warning. This is caused by Symfony itself, see https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Routing/Route.php#L307

Guite:
@craigh http://symfony.com/blog/symfony-2-7-1-released

`bug #14900 Silence deprecation warnings by default (reecefowell)`

craigh:
see comment in slack.

weaverryan:
I'm not convinced that the code in [Route.php](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Routing/Route.php#L307) is responsible for the deprecated warnings about `_method`. `_method` *is* being added to the `requirements` property, but `sanitizeRequirement()` is never called for it (`sanitizeRequirement()` is only called when you actually add a requirement via the public method). Even if you serialized the route and deserialized it, you'd end up with the same `Route` object, still without `sanitizeRequirement()` bing called on the `_method`.

I don't know how the `zikula_routes_route` table is populated exactly, but if a Symfony `Route` object is used to populate it, then later the data in `zikula_routes_route` is used to re-create a Symfony `Route` object, then you probably would be calling `Route::setRequirements()`, which would trigger the warning. But if that's the case, you should be able to filter the _method out before it's saved into the database.

I hope that helps!

Guite:
Thanks @weaverryan for your explaination.
I guess that https://github.com/zikula/core/blob/1.4/src/system/ZikulaRoutesModule/Routing/RouteLoader.php#L85 could be the problem then.

@cmfcmf what do you think?

cmfcmf:
Yes @Guite, this probably is it.

cmfcmf:
refs
- https://github.com/schmittjoh/JMSI18nRoutingBundle/pull/152
- https://github.com/Elao/WebProfilerExtraBundle/issues/64

cmfcmf:
The thing @weaverryan suggested has now bin added in #2498.

craigh:
THANK YOU @weaverryan for taking the time to help us! (conversation began [here](http://knpuniversity.com/blog/upgrading-symfony-2.7#comment-2077695979))

weaverryan:
Cheers! Happy to be able to help out another library in the ecosystem :)

PR #2505, opened 2015-06-21
[WIP] Travis update

cmfcmf:
This does make the Travis Builds a bit more useful again. Depends on https://github.com/matthiasnoback/symfony-console-form/pull/7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: php
-sudo: false
 
 php:
   - 5.4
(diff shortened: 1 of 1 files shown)


craigh:
wow cool. I think this might be ok to include in 1.4.0 because it isn't technically part of the core code. Even though it removes the php 5.3 test, IMO, that is OK. Maybe others feel differently.

cmfcmf:
Well, the PHP 5.3 test doesn't help as it always fails :-D

craigh:
yes, exactly. But that's my fault for including the `vierbergenlars/php-semver` lib.

PR #2542, opened 2015-07-18
formatForOS() corrupts path beginning with "../"

rmaiwald:
A relative path beginning with "../" results in an array_pop() on an empty array (stack underflow). In a relative path a leading ".." should be left in place. A leading ".." should be removed only in an absolue path.

### DIFF
--- a/src/lib/util/DataUtil.php
+++ b/src/lib/util/DataUtil.php
@@ -380,9 +380,13 @@ class DataUtil
             if ($current == '.') {
                 // current path element is a dot, so we don't do anything
             } elseif ($current == '..') {
-                // current path element is .., so we remove the last path in case of relative paths
-                if (!$absolutepathused) {
+                // current path element is .., so we remove the last path
+                if (count($clean_array) > 0) {
                     array_pop($clean_array);
+                } else {
+                    if (!$absolutepathused) {
+                        array_push($clean_array, $current);
+                    }
                 }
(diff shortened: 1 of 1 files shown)


craigh:
@rmaiwald thank you for the PR!

We will look into  including this possibly in 1.4.1 as 1.4.0 is already preparing for release.

If possible, can you provide the use case where this became a problem for you?  thank you!

rmaiwald:
Well, my use case is, that the directory 'userdata' is not a subdirectory of the Zikula software package on out web server but ist located beside it. The directory layout looks like this:

/zk (the Zikula software package)
/zk.tmp (service page when Zikula is offline)
/userdata

With this directory layout all relative paths to userdata begin with "../userdata".

The idea behind this directory layout is, that I can upload a new Zikula software version in a new subdirectory, e.g. "/zk13". And I do not have to download about 6 GB of userdata from /zk/userdata and upload them to /zk13/userdata via FTP.

craigh:
@Guite @cmfcmf thoughts?

craigh:
@rmaiwald thank you for your contribution, but I don't think we can allow this type of behavior due to security concerns. Please see this discussion:

refs https://github.com/zikula/core/pull/458

(the "ghost" in the discussion is former lead-developer, drak)

thank you again. I hope you will continue to use and contribute to Zikula!

PR #2563, opened 2015-08-17
Convert system and themes to PSR-4

craigh:
Convert all system modules and all system themes to PSR-4. maintain BC.

## REQUIRES COMPOSER UPDATE !!!

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | yes*
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | #2424
| License           | MIT
| Doc PR            | -
| Changelog updated | yes

* deprecated items are only a result of namespace changes and a BC layer has been added to compensate.

Issue #2500, opened 2015-06-17
2.0.0 Extension Model/Standard


craigh:
This ticket is meant to encourage a discussion and (eventual) finalization of a model/standard/prototype/pattern for Core 2.0.0 Extension Bundles.

current 1.4.0 standard is [documented here](https://github.com/zikula/core/blob/1.4/src/docs/en/dev/UPGRADE-1.4.0.md#modulestructure)

Core 2.0.0 will 'evolve' this standard but is **not required** to maintain Backward Compatibility.

Easy agreement on the following:
 - Extends `\Zikula\Core\AbstractBundle` which extends `\Symfony\Component\HttpKernel\Bundle\Bundle`
   - This implies certain Bundle functionality as a Symfony Bundle.
 - PSR-4 required (PSR-0 is deprecated)
   - requires a `composer.json` file ([1.4 Module Standard](https://github.com/zikula/core/blob/1.4/src/docs/en/dev/UPGRADE-1.4.0.md#module-composerjson) & [1.4 Theme Standard](https://github.com/zikula/core/blob/1.4/src/docs/en/dev/UPGRADE-1.4.0.md#theme-composerjson))

craigh:
Current questions in my mind:

#### Version
 - I think we can move this data to the `composer.json` file
   - how/where?
 - HookBundle Setup might be moved to [Service Container tags](http://symfony.com/doc/current/book/service_container.html#tags) with a compiler pass to gather them.

#### Installer
 - Create an Interface instead of an Abstract class.
```php
<?php
interface ExtensionInstaller
{
    public function setBundle(Bundle $bundle);
    public function install();
    public function upgrade($oldVersion);
    public function remove();
}
```
 - Most Installer implementations will also implement the `ContainerAwareInterface`, although, the container can currently be extracted from the Bundle also. Discussion here, I suppose on best way to handle this.

#### Bundle class
 - could expand this to directly set Installer class and extract Version info from composer file

#### Hooks
 - I'm wondering if it is possible to use *Service Container Tags* (see above) for all hooks instead of persisting them.

#### AdminInterface
 - Purpose: https://github.com/zikula/core/issues/2112
 - Idea: [AdminAuthInterface](https://github.com/zikula-modules/Pages/blob/master/AdminAuthInterface.php) / [AdminAuthListener](https://github.com/zikula-modules/Pages/blob/master/Listener/AdminAuthListener.php)

#### Zikula\Core\Api\AbstractApi
[...]

PR #2564, opened 2015-08-20
[WIP] Theme overrides

craigh:
## Merge after #2563
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | #2500
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/system/RoutesModule/Resources/views/Route/view.tpl
+++ b/src/system/RoutesModule/Resources/views/Route/view.tpl
@@ -8,39 +8,45 @@
 {if count($items) > 1}
 {pageaddvar value='jquery-ui' name='javascript'}
 <script type="text/javascript">
-    jQuery(function(){
-        // Return a helper with preserved width of cells
-        var fixHelper = function(e, ui) {
-            ui.children().each(function() {
-                jQuery(this).css({width: jQuery(this).width()});
-            });
-            return ui;
-        };
+    /* <![CDATA[ */
+    ( function() {
+        jQuery(document).ready(function() {
+            jQuery(function(){
+                // Return a helper with preserved width of cells
+                var fixHelper = function(e, ui) {
+                    ui.children().each(function() {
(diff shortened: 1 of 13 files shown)


src/system/RoutesModule/Resources/views/Route/view.tpl
```diff
@@ -8,39 +8,45 @@
...
-    jQuery(function(){
-        // Return a helper with preserved width of cells
-        var fixHelper = function(e, ui) {
-            ui.children().each(function() {
-                jQuery(this).css({width: jQuery(this).width()});
-            });
-            return ui;
-        };
+    /* <![CDATA[ */
+    ( function() {
```
cmfcmf: If you wrap it into an anonymus function, you could have also replaced `jQuery` by `$`.
```js
(function ($) {
    // ...
})(jQuery)
```

craigh: IANAJSP (I Am Not A JavaScript Programmer)

but the way I understand it , doing it your way just defines `$ = jQuery` and allows it to be used as such. I figured the way I did it just stuck with the `jQuery` var name. no?

cmfcmf: See, I was confused why you wrapped the code. Because commonly, I wrap it to be able to use `$` as `jQuery` inside it.

IANAJSPAWBKAFT (I Am Not A JavaScript Programmer As Well But Know A Few Tricks)

craigh: well, I'm not sure. maybe @jusuff knows and is willing to comment, because HIAJSP.

PR #2569, opened 2015-08-27
Theme override by annotation

craigh:
implement annotation reader and change themes based on `@Admin` annotation

@cmfcmf @Guite your thoughts?


src/lib/Zikula/Bundle/CoreBundle/EventListener/ThemeListener.php
```diff
@@ -161,16 +164,63 @@ public function setUpThemePathOverrides(FilterControllerEvent $event)
...
+        $this->changeThemeByAnnotation($controller, $method);
+    }
+
+    /**
+     * Change a theme based on the annotation
+     * @param $controller
+     * @param $method
+     * @return array|bool|string
+     */
+    public function changeThemeByAnnotation($controller, $method)
```
craigh: was considering if I should move this method to the theme engine...

Guite: I think rather not, because it changes the theme which looks like an operation from outside... @cmfcmf what do you mean?

cmfcmf: I don't know.

src/lib/Zikula/Core/Theme/Annotation/Admin.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Zikula\Core\Theme\Annotation;
+
+use Doctrine\Common\Annotations\Annotation;
+
+/**
+ * @Annotation
+ * Class Admin
+ * @package Zikula\Core\Theme\Annotation
+ */
+class Admin extends Annotation
```
cmfcmf: As said in my other comment, I would like this to be called `AdminTheme`.

src/lib/Zikula/Bundle/CoreBundle/EventListener/ThemeListener.php
craigh: @Guite I don't follow what you mean. After thinking about it more, I think I should move it for testability (future). Also wondering about moving the Listener into the `Zikula/Core/Theme/Listener` directory (instead of `Zikula/Core/Bundle/CoreBundle/Listener`). opinion?

Guite: I've no real opinion about this. Maybe we need a general distinction of what comes into the core bundle and what not. My feeling says put it into the bundle but I have no reasoning.

craigh: re: moving the Listener class - looking through Symfony, typically the EventListeners are stored in the Bundles (3 of 5 Bundles have a directory). Components typically do *not* have an EventListener directory but an exception is the HttpKernel Component which has many Listeners.

The Zikula/Core isn't organized as **Components** specifically so the comparison isn't 100% true. It is just simply a collection of classes.

The reason I would consider moving the class is to further encapsulate (I'm not sure if that is the proper use of this term here) the Theme classes into one directory.

Issue #2494, opened 2015-06-13
reconsider AbstractApi (and getLinks)


craigh:
temporary solution of AbstractApi may not be best solution for Core 2.0.0

reconsider possible solutions.

/cc @cmfcmf

craigh:
@cmfcmf says: I think this is the way to go:

http://symfony.com/doc/current/book/service_container.html#book-service-container-tags

the tag name would be `zikula.getlinks` or something

Christian Flach [1:41 PM]
So everyone who wants to provide links has to define a service and tag it.

craigh:
then he said

> I thougt about the service-tag thing again and I think an event is really better in the end.

and

> right, now I remember. We could also use an event. I don't know what's better.
> probably the event

craigh:
closed by #2575

Issue #1950, opened 2014-08-23
API vs. Util vs. services 


craigh:
I think we need to give extension developers more guidance on the purpose of API and better define an extension architecture to include Util classes or something like that.

In the past, we used module API functions as (somewhat) easily accessible "helper" functions to do the heavy lifting we wanted to keep out of the controller - database interaction, utility function and just a general "catch all" for code we don't want elsewhere. In addition, the term "API" has been misused here. If something is a "API" it might be assumed that the function is accessible from outside the component [see def](http://en.wikipedia.org/wiki/Application_programming_interface) and [ref News ticket](https://github.com/zikula-modules/News/issues/136)

As we move toward Symfony with Doctrine, the traditional API function seems less useful and relevant. Of course, there is nothing forcing a developer to use the published structure and utilizing API type classes except for the ability to call `ModUtil::apiFunc()` which IMO is becoming much clumsier than importing a Util class and issuing `MyModUtil::myFunc()`.

I guess I'm suggesting we do away with the traditional understanding of API classes and move to more Util classes and only use API for functions intended to be externally accessible.

thoughts/opinions?

Guite:
Closing as fixed, since the 2.0 spec actively supports the usage of services.

PR #2575, opened 2015-09-01
remove the Zikula\Core\Api\AbstractApi class. 

craigh:
It is a BC Break. But since it was just introduced and there are very unlikely any modules using it yet, I believe it to be safe to remove.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | YES
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2494
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | YES

@cmfcmf @Guite @phaidon @Kaik please vote :+1: :-1:  (others welcome also of course)


Guite:
What is the replacement?

I have already 4 modules using it.
Michael just started with the 1.4 version of Eternizer https://github.com/zikula-modules/Eternizer/ which also uses it now.

Guite:
Just saw that the new `getLinks` service can be used as replacement.
So personally I am fine with this change :+1: 

@portugao could you wait with the Eternizer release until 1.4.1 is there? The change is a rather small one.

Portugao:
Yes.

craigh:
Replacements: yes, the LinkContainer and getLinks service are intended replacements for the link Api in <=1.4.0. Api classes are not required at all but of course could still be used and identified as a service instead of extending a base class. This would be the preference. refs #1950 

another option of course would be to simply `@deprecate` the entire class and leave it there until 2.0 but I don't want to encourage its usage, so I thought removing it was the better solution.

Portugao:
:-1: 
Would prefere to @deprecate the class and leave it until 2.0.

craigh:
> There is enough to understand with 1.4.

that is why I am suggesting eliminating it. :smile: don't learn the 1.4 way and have to learn the 1.4.1/2.0 way again. just jump straight to the latter.

Kaik:
:+1: I think that 1.4.1 should introduce much more BC so we can have quicker transition, at least at this kind of "easy level" as getLinks. You can still use Api class for your own things just this "internal stuff" is moved. Fine with me. :+1:

craigh:
@Kaik I'm not sure you understand. This "new" AbstractApi was introduced in Core-1.4 but will NOT be used in Core-2.0. (It probably shouldn't have been introduced - my fault). So this is just to remove it. The FC-layer with the new getLInks service is implemented already and will be the way it will work in Core-2.0. I hope that is clearer.

Kaik:
@craigh I thought that AbstractApi will become what Api is... in general so other application/module can communicate with "this module". Maybe not that big like google api but kind of similar exposing some default functionality between modules. I do not have problem with any of ways anyway as I was using Api only for links in my modules.

Guite:
@Kaik Api classes are going to be replaced by Symfony services: http://symfony.com/doc/current/book/service_container.html
About communication: by tagging services http://symfony.com/doc/current/book/service_container.html#tags event listeners can interact with them.

craigh:
> I thought that AbstractApi will become what Api is...

yes, modules are allowed to publish a 'public' Api if they wish, but I don't want the core to provide an Abstract to base this on. The providing module will need to declare the Api a symfony service and inject whatever services it requires there.

Kaik:
> but I don't want the core to provide an Abstract to base this on

Well for me it is ok. Like I said I'm using it only for getlinks. There is nothing else to put in it since symfony/doctrine everything what used to be there is now in repository

craigh:
So it looks like we have 3 :+1: and 1 :-1: 

@cmfcmf ? @phaidon ?

cmfcmf:
Yes, remove it please.

Guite:
https://github.com/Guite/MostGenerator/commit/c2447deffb6ac2cd561c6cb3595b083f32ccec10

PR #2576, opened 2015-09-04
Don't save all the routes into the database.

cmfcmf:
I got it almost working now. It now auto-reloads and detects any new or changed routes when in dev mode. Reloading also is much faster. I haven't written upgrade code nor checked the installer yet. Also, some dead code needs to be removed and the Web UI to add Routes has to be checked and improved.

- [x] changelog
- [x] implement upgrade code
- [x] change the db
- [x] better web ui
- [x] check core install (checked at https://github.com/zikula/core/commit/4b5100e26ce911194b7ba619005857a188622ea2)
- [x] check core upgrade (checked at (https://github.com/zikula/core/commit/4b5100e26ce911194b7ba619005857a188622ea2)


craigh:
this is great! Another thing to check on will be whether the "second pass" is still required during module installation. It would be terrific if we can remove it.

src/system/RoutesModule/Api/Base/CacheApi.php
```diff
@@ -14,14 +14,14 @@
...
-use Zikula_AbstractApi;
+use Zikula_AbstractBase;
 use Zikula_View;
 use Zikula_View_Theme;
 
 /**
  * Cache api base class.
  */
-class CacheApi extends Zikula_AbstractApi
+class CacheApi extends Zikula_AbstractBase
```
craigh: to me this seems odd. We've never extended directly from AbstractBase. It is a deprecated class anyway. if you are going to use a deprecated class, why not just use AbstractApi like before? If not that then why use anything?

cmfcmf: This is what the MOST generator produced.

craigh: ping @Guite thoughts?

Guite: This is subject of transition. Util classes also extend from AbstractBase. The difference between AbstractBase and AbstractApi is that AbstractApi is removed already.

craigh: I think you are confusing legacy `Zikula_AbstractApi` and `Zikula\Core\Api\AbstractApi` which was [just removed](https://github.com/zikula/core/pull/2575). The legacy [Zikula_AbstractApi](https://github.com/zikula/core/blob/1.4/src/lib/legacy/Zikula/AbstractApi.php) class still exists.

cmfcmf:
Note to self: http://symfony.com/doc/current/components/routing/introduction.html#using-prefixes

.gitignore
```diff
@@ -19,6 +19,7 @@
 !/src/themes/PrinterTheme
 !/src/themes/RssTheme
 !/src/themes/SeaBreezeTheme
+!/src/themes/MockTheme
```
craigh: it appears that this branch/PR has some old commits that were removed from the git history of the core. rebasing _may_ remove these but if not, you will have to do something different. both commits related to the MockTheme solution must be removed.

src/system/RoutesModule/Api/Base/CacheApi.php
craigh: I still think this is a wrong use of `AbstractBase`, by the way. I don't think this class was ever intended to be used outside of being extended by `Zikula_AbstractApi` and `Zikula_AbstractController` (both obviously legacy classes). I strong recommend discontinuing its use outside the core and refactoring to a service class. If you must use a legacy class, use `Zikula_AbstractApi` directly.

craigh:
@cmfcmf 
> Note to self: http://symfony.com/doc/current/components/routing/introduction.html#using-prefixes

please look into that. We may be solving this problem (prefixing routes) the incorrect way. this may be much easier.

craigh:
@cmfcmf the problem with that solution is that you cannot 'opt out' any give route like you have allowed with the `_zkNoPrefix` param. so maybe that is not a good solution? or ... maybe you work with TWO collections at once, one WITH the prefix and one without...

cmfcmf:
> @cmfcmf the problem with that solution is that you cannot 'opt out' any give route like you have allowed with the _zkNoPrefix param. so maybe that is not a good solution? or ... maybe you work with TWO collections at once, one WITH the prefix and one without...

Right, that was the reason why I burried the idea of using addPrefix. Internally, it doesn't do any magic: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Routing/RouteCollection.php#L150

cmfcmf:
While I was looking at the RouteCollection class right now, I discovered that there is a RouteCollectionBuilder. Worth checking out I think: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Routing/RouteCollectionBuilder.php

craigh:
it may not do any magic, but it might eliminate the need in `\Zikula\RoutesModule\Translation\ZikulaPatternGenerationStrategy` and `\Zikula\RoutesModule\Translation\ZikulaRouteTranslationExtractor` but maybe not....

cmfcmf:
> it may not do any magic, but it might eliminate the need in \Zikula\RoutesModule\Translation\ZikulaPatternGenerationStrategy and \Zikula\RoutesModule\Translation\ZikulaRouteTranslationExtractor but maybe not....

I don't think it does, but I might be wrong of course :boat:

craigh:
I've just found that the `.generated` files that have anything to do with Doctrine will confuse the Doctrine system such that `php app/console doctrine:schema:validate` produces errors

PR #2580, opened 2015-09-07
add method to configure twig-based theme variables (form-abstraction)

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -

### DIFF
--- a/src/lib/Zikula/Core/AbstractTheme.php
+++ b/src/lib/Zikula/Core/AbstractTheme.php
@@ -77,6 +77,51 @@ abstract class AbstractTheme extends AbstractBundle
     }
 
     /**
+     * load the themevars into the themeEngine global vars
+     */
+    public function setThemeVars()
+    {
+        $this->getContainer()->get('zikula_core.common.theme.themevars')->replace($this->getThemeVars());
+    }
+
+    /**
+     * Get the theme variables from both the DB and the .yml file.
+     * @return array|string
+     */
+    public function getThemeVars()
+    {
+        $dbVars = \ModUtil::getVar($this->name);
+        $defaultVars = $this->getDefaultThemeVars();
+        $combinedVars = array_merge($defaultVars, $dbVars);
+        if (array_keys($dbVars) != array_keys($combinedVars)) {
+            // First load of file or vars have been added to the .yml file.
+            \ModUtil::setVars($this->name, $combinedVars);
+        }
+
+        return $combinedVars;
+    }
+
+    /**
+     * Get the default values from variables.yml.
+     * @return array
+     */
+    public function getDefaultThemeVars()
+    {
+        $defaultVars = [];
+        $themeVarsPath = $this->getConfigPath() . '/variables.yml';
+        if (file_exists($themeVarsPath)) {
+            if ($this->getContainer()) {
+                $yamlVars = Yaml::parse($themeVarsPath);
+                foreach($yamlVars as $name => $definition) {
+                    $defaultVars[$name] = $definition['default_value'];
+                }
+            }
+        }
+
+        return $defaultVars;
+    }
+
+    /**
      * Is theme twig (Core-2.0) based?
      * @deprecated
      * @return bool
--- a/src/lib/Zikula/Core/AbstractTheme.php
+++ b/src/lib/Zikula/Core/AbstractTheme.php
@@ -79,7 +79,7 @@ abstract class AbstractTheme extends AbstractBundle
     /**
      * load the themevars into the themeEngine global vars
      */
-    public function setThemeVars()
+    public function loadThemeVars()
     {
         $this->getContainer()->get('zikula_core.common.theme.themevars')->replace($this->getThemeVars());
     }
@@ -111,7 +111,7 @@ abstract class AbstractTheme extends AbstractBundle
         $themeVarsPath = $this->getConfigPath() . '/variables.yml';
         if (file_exists($themeVarsPath)) {
             if ($this->getContainer()) {
-                $yamlVars = Yaml::parse($themeVarsPath);
+                $yamlVars = Yaml::parse(file_get_contents($themeVarsPath));
                 foreach($yamlVars as $name => $definition) {
                     $defaultVars[$name] = $definition['default_value'];
                 }
--- a/src/system/ThemeModule/Controller/VarController.php
+++ b/src/system/ThemeModule/Controller/VarController.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * Copyright Zikula Foundation 2015 - Zikula Application Framework
+ *
+ * This work is contributed to the Zikula Foundation under one or more
+ * Contributor Agreements and licensed to You under the following license:
+ *
+ * @license GNU/LGPLv3 (or at your option, any later version).
+ *
+ * Please see the NOTICE file distributed with this source code for further
+ * information regarding copyright and licensing.
+ */
+
+namespace Zikula\ThemeModule\Controller;
+
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route; // used in annotations - do not remove
+use Symfony\Component\Yaml\Yaml;
+use Symfony\Component\HttpFoundation\Request;
+use Zikula\Core\Controller\AbstractController;
+use Zikula\Core\Theme\Annotation\Theme; // used in annotations - do not remove
+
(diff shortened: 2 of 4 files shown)


craigh:
ping @cmfcmf this is the method I was referring to. see [this commit](https://github.com/zikula/core/commit/b5b3ca4bf9929d26d8a9af2c29242440a53b8e53) for sample of variables.ml file

src/lib/Zikula/Core/AbstractTheme.php
```diff
@@ -77,6 +77,51 @@ public function generateThemedBlock(array $block)
     }
 
     /**
+     * load the themevars into the themeEngine global vars
+     */
+    public function setThemeVars()
```
cmfcmf: I don't like the name too much - *setting* a *var* in Zikula context always meant a function like `setVar($name, $value)`. I'd rather change this to `initThemeVars` or `setThemeVarsInContainer` or something.

src/lib/Zikula/Core/AbstractTheme.php
```diff
@@ -77,6 +77,51 @@ public function generateThemedBlock(array $block)
...
+     * Get the default values from variables.yml.
+     * @return array
+     */
+    public function getDefaultThemeVars()
+    {
+        $defaultVars = [];
+        $themeVarsPath = $this->getConfigPath() . '/variables.yml';
+        if (file_exists($themeVarsPath)) {
+            if ($this->getContainer()) {
+                $yamlVars = Yaml::parse($themeVarsPath);
```
cmfcmf: This shold be changed to ` Yaml::parse(file_get_contents('/path/to/file.yml'));`
> Because it is currently possible to pass a filename to this method, you must validate the input first. Passing a filename is deprecated in Symfony 2.2, and will be removed in Symfony 3.0.
-- http://symfony.com/doc/current/components/yaml/introduction.html

src/system/ThemeModule/Controller/VarController.php
```diff
@@ -0,0 +1,84 @@
...
+use Symfony\Component\Yaml\Yaml;
+use Symfony\Component\HttpFoundation\Request;
+use Zikula\Core\Controller\AbstractController;
+use Zikula\Core\Theme\Annotation\Theme; // used in annotations - do not remove
+
+class VarController extends AbstractController
+{
+    /**
+     * Configure a theme's variables based on provided .yml definitions for each field.
+     * @Route("/admin/var/{themeName}")
```
cmfcmf: `/admin/variables/{themename}` would be a bit more verbose.

cmfcmf: Nevermind, if I read the next line before commenting, I would know better :boat:

src/system/ThemeModule/Controller/VarController.php
```diff
@@ -0,0 +1,84 @@
...
+     * @param Request $request
+     * @param string $themeName
+     * @return mixed
+     * @throws \InvalidArgumentException if theme type is not twig-based
+     */
+    public function varAction(Request $request, $themeName)
+    {
+        $themeBundle = \ThemeUtil::getTheme($themeName);
+        if (!$themeBundle->isTwigBased()) {
+            throw new \InvalidArgumentException('Theme type must be twig-based in ' . __FILE__ . ' at line ' . __LINE__ . '.');
```
cmfcmf: Should maybe be `HttpNotFoundException`.

src/system/ThemeModule/Controller/VarController.php
```diff
@@ -0,0 +1,84 @@
...
+        }
+        $themeVarsPath = $themeBundle->getConfigPath() . '/variables.yml';
+        if (file_exists($themeVarsPath)) {
+            $variableDefinitions = Yaml::parse($themeVarsPath);
```
cmfcmf: same with the Yaml::parse here.

src/system/ThemeModule/Controller/VarController.php
```diff
@@ -0,0 +1,84 @@
...
+            /** @var \Symfony\Component\Form\FormBuilder $formBuilder */
+            $formBuilder = $this->createFormBuilder($themeBundle->getThemeVars());
+            foreach ($variableDefinitions as $fieldName => $definitions) {
+                $options = isset($definitions['options']) ? $definitions['options'] : [];
+                if (isset($definitions['type'])) {
+                    $formBuilder->add($fieldName, $definitions['type'], $options);
+                }
+            }
+            $formBuilder->add('save', 'submit', array('label' => 'Save'))
```
cmfcmf: Missing `__()`?

src/system/ThemeModule/Controller/VarController.php
```diff
@@ -0,0 +1,84 @@
...
+            $form = $formBuilder->getForm();
+
+            $form->handleRequest($request);
+            if ($form->isValid()) {
+                if ($form->get('save')->isClicked()) {
+                    // pseudo-hack to save theme vars in to modvars table
+                    \ModUtil::setVars($themeName, $form->getData());
+                    $this->addFlash('status', __('Done! Theme configuration updated.'));
+                }
+                if ($form->get('toDefault')->isClicked()) {
```
cmfcmf: I'd prefer elseif's here.

src/system/ThemeModule/Controller/VarController.php
```diff
@@ -0,0 +1,84 @@
...
+     * @throws \InvalidArgumentException if theme type is not twig-based
+     */
+    public function varAction(Request $request, $themeName)
+    {
+        $themeBundle = \ThemeUtil::getTheme($themeName);
+        if (!$themeBundle->isTwigBased()) {
+            throw new \InvalidArgumentException('Theme type must be twig-based in ' . __FILE__ . ' at line ' . __LINE__ . '.');
+        }
+        $themeVarsPath = $themeBundle->getConfigPath() . '/variables.yml';
+        if (file_exists($themeVarsPath)) {
```
cmfcmf: Instead of wrapping all the stuff inside the if, rather invert the condition to decrease indenting.

src/system/ThemeModule/Controller/VarController.php
craigh: it's an argument.

src/system/ThemeModule/Resources/views/Admin/var.html.twig
```diff
@@ -0,0 +1,18 @@
+{% form_theme form with [
+'ZikulaFormExtensionBundle:Form:bootstrap_3_zikula_admin_layout.html.twig',
+'ZikulaFormExtensionBundle:Form:form_div_layout.html.twig'
```
cmfcmf: What does the second form layout do?

cmfcmf:
I like it. The only downside I found is the following: The labels of the variables aren't translatable (or are they?), but IMHO this is not too important. Theme variables are more advanced anyways.

src/system/ThemeModule/Controller/VarController.php
cmfcmf: Yes, but it also is a controller. I expect a *404 Not Found* when opening an URL like `/admin/var/IDontExist` and not a *500 Internal Server Error*, no?

craigh:
> The labels of the variables aren't translatable (or are they?)

no they are not. I don't believe this is possible.

src/system/ThemeModule/Resources/views/Admin/var.html.twig
craigh: allows an 'icon' parameter (adding usage now)

cmfcmf: Would it also be possible to allow a "helptext" parameter to display a little explanation below each setting? (Using Bootstrap's `help-block`?)

craigh: with some work I believe so. I cannot remember but I don't think this is natively supported (in Forms). But I remember seeing some implementations of this online. It would require a similar override of the `form_div_layout.html.twig` template. I would prefer to not implement that here, but could be considered for later.

cmfcmf: OK :+1:

PR #2583, opened 2015-09-09
[1.3] Load correct theme in case Mobile theme is enabled and Mobile detect is enabled

nmpetkov:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| 1.4 PR        | -
| Refs tickets  | -
| License       | MIT
| Doc PR        | -

This PR is related to Zikula 1.3 branch only, and corrects a bug in default theme loading in case Mobile theme is enabled and Mobile detect is also enabled, but client is not mobile.
In particular, if described conditions occurs, then "Printer" and "RSS" themes, passed in URL, will not load.

### DIFF
--- a/src/lib/util/UserUtil.php
+++ b/src/lib/util/UserUtil.php
@@ -1906,6 +1906,8 @@ class UserUtil
             $detect = new Mobile_Detect();
             if ($detect->isMobile()) {
                 $pagetheme = 'Mobile';
+`            } else {
+                $pagetheme = FormUtil::getPassedValue('theme', null, 'GETPOST');
             }
         } else {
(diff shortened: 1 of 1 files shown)


craigh:
should have asked for a changelog entry before merging. can you please submit another PR?

also - you should be aware that there *may* not actually be a 1.3.11 release... not sure.

nmpetkov:
Yes, I also hope 1.4 branch soon to become so stable as 1.3 branch :-)

PR #2594, opened 2015-09-16
[WIP] Make thumbnailing a transformation, so that other transformations can be applied after thumbnailing

cmfcmf:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | yes, in edge cases
| Deprecations?     | no
| Tests pass?       | ---
| Fixed tickets     | ---
| Refs tickets      | ---
| License           | MIT
| Doc PR            | ---
| Changelog updated | yes

This makes it possible to use the Imagine plugin to create thumbnails and apply further transformations after creating the thumbnail. Currently, all custom transformations are applied before creating the thumbnail. With this PR, it is possible to add transformations which are applied after creating the thumbnail. I.e. watermark the image's thumbnail.

This is a minor BC break for modules using Imagine transformations with a priority greater than 50. I am not aware of any modules doing so. If there are such modules, the fix is as trivial as changing the priorities of the custom transformations.

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -5,6 +5,8 @@ CHANGELOG - ZIKULA 1.4.x
 
  - BC Breaks:
      - Removed `Zikula\Core\Api\AbstractApi` that was introduced only in 1.4.0 (#2494)
+     - If you use the Imagine System plugin and add custom transformations with a priority greater than 50,
+       these are now applied *after* the thumbnail is generated. (#2594)
  - Deprecated:
      - Twig function `pageAddVar()` deprecated. Use `pageAddAsset()` or `pageSetVar()` instead.
  - Fixes:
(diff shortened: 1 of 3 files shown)


cmfcmf:
I would like your feedback on it, esspecially whether you know of any modules using transformations and if you think this tiny BC break is ok for 1.4.1.

rallek:
Is news creating thumbnails that way?

cmfcmf:
I don't think so: https://github.com/zikula-modules/News/issues/18

And even if it did, it would probably not use any custom transformations and therefore not be affected by this PR.

Guite:
While all MOST modules (containing upload fields) use Imagine I do not know any single module utilising custom transformations.

cmfcmf:
Once the Theme and Asset issues are fixed, I will be able to test this once again and verify that it works. Apart from that, it should be ready.

craigh:
@cmfcmf please rebase this PR and also add a not in the BC Break area of the changelog related to potential issues with this. Test and merge if you are ready. I think the Theme and Asset issues you mentioned are fixed?

cmfcmf:
> add a not in the BC Break area of the changelog

Done and tested. Going to merge this once the CI finishes.

PR #2597, opened 2015-09-17
Include in the {ext} function filtering for CLIP

trichers:
To use the filter-function in the generated link


craigh:
can you please add a changelog entry for this before merging? thanks

trichers:
I thought merging means it is a proposal?

craigh:
> I thought merging means it is a proposal?

this PR is the proposal. With it, I would like you to include an update to the changelog as well. _Merging_ means to combine your code into the actual codebase.

Issue #2604, opened 2015-09-20
Calling clear_cacheid_allthemes generates error after Theme Refactoring


cmfcmf:
MOST calls 
```php
        $theme = Zikula_View_Theme::getInstance();
        $theme->clear_cacheid_allthemes($cacheIds);
```
in `Api/Base/CacheApi` (see Routes Module for example, try to edit, not create, a route) which results in 
```
[1] InvalidArgumentException: Invalid arguments array received
    at n/a
        in E:\wamp\www\core\src\system\ThemeModule\Api\UserApi.php line 250

    at Zikula\ThemeModule\Api\UserApi->getpageconfiguration(array('theme' => 'ZikulaBootstrapTheme', 'filename' => null))
        in  line 

    at call_user_func(array(object(UserApi), 'getpageconfiguration'), array('theme' => 'ZikulaBootstrapTheme', 'filename' => null))
        in E:\wamp\www\core\src\lib\util\ModUtil.php line 1187

    at ModUtil::exec('ZikulaThemeModule', 'user', 'getpageconfiguration', array('theme' => 'ZikulaBootstrapTheme', 'filename' => null), true, null)
        in E:\wamp\www\core\src\lib\util\ModUtil.php line 1263

    at ModUtil::apiFunc('ZikulaThemeModule', 'user', 'getpageconfiguration', array('theme' => 'ZikulaBootstrapTheme', 'filename' => null))
        in E:\wamp\www\core\src\lib\legacy\Zikula\View\Theme.php line 787

    at Zikula_View_Theme->load_config()
        in E:\wamp\www\core\src\lib\legacy\Zikula\View\Theme.php line 231

    at Zikula_View_Theme->__construct(object(appDevDebugProjectContainer), 'ZikulaBootstrapTheme')
        in E:\wamp\www\core\src\lib\legacy\Zikula\View\Theme.php line 295

    at Zikula_View_Theme::getInstance()
        in E:\wamp\www\core\src\system\RoutesModule\Api\Base\CacheApi.php line 95

    at Zikula\RoutesModule\Api\Base\CacheApi->clearItemCache(array('ot' => 'route', 'item' => object(RouteEntity)))
        in  line 

    at call_user_func(array(object(CacheApi), 'clearItemCache'), array('ot' => 'route', 'item' => object(RouteEntity)))
        in E:\wamp\www\core\src\lib\util\ModUtil.php line 1187

[...]

PR #2605, opened 2015-09-21
create MockTheme class to use when Theme is Twig-based. refs #2604

craigh:


craigh:
I did some other testing and this works for me. If there are problems then we will fix them additionally.

rallek:
What is this doing? Is it somehow usefull for someone like me?

Guite:
It is just a temporary thing for backwards compatibility. Nothing to be used by end users.

PR #2624, opened 2015-10-04
Add ZikulaExtensionModule/VariableApi

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | yes
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | yes

Add ZikulaExtensionModule/VariableApi to manage System, Module and Theme variables. Add tests and documentation. update changelog.

### DIFF
--- a/src/lib/util/ModUtil.php
+++ b/src/lib/util/ModUtil.php
@@ -115,8 +115,8 @@ class ModUtil
 
         // This loads all module variables into the modvars static class variable.
         $em = ServiceUtil::get('doctrine.entitymanager');
-        /** @var \Zikula\Core\Doctrine\Entity\ExtensionVarEntity[] $modvars */
-        $modvars = $em->getRepository('Zikula\Core\Doctrine\Entity\ExtensionVarEntity')->findAll();
+        /** @var \Zikula\ExtensionsModule\Entity\ExtensionVarEntity[] $modvars */
+        $modvars = $em->getRepository('Zikula\ExtensionsModule\Entity\ExtensionVarEntity')->findAll();
         foreach ($modvars as $var) {
             if (!array_key_exists($var->getModname(), self::$modvars)) {
                 self::$modvars[$var->getModname()] = array();
@@ -165,6 +165,9 @@ class ModUtil
 
     /**
      * Checks to see if a module variable is set.
+     * @deprecated at Core-1.4.1
+     * @see \Zikula\ExtensionsModule\Api\VariableApi::has()
+     * @see service zikula_extensions_module.api.variable
      *
      * @param string $modname The name of the module.
      * @param string $name    The name of the variable.
@@ -196,6 +199,10 @@ class ModUtil
 
     /**
      * The getVar method gets a module variable.
+     * @deprecated at Core-1.4.1
+     * @see \Zikula\ExtensionsModule\Api\VariableApi::get()
+     * @see \Zikula\ExtensionsModule\Api\VariableApi::getAll()
+     * @see service zikula_extensions_module.api.variable
      *
      * If the name parameter is included then method returns the
      * module variable value.
@@ -254,6 +261,9 @@ class ModUtil
 
     /**
      * The setVar method sets a module variable.
+     * @deprecated at Core-1.4.1
+     * @see \Zikula\ExtensionsModule\Api\VariableApi::set()
+     * @see service zikula_extensions_module.api.variable
      *
      * @param string $modname The name of the module.
      * @param string $name    The name of the variable.
@@ -273,14 +283,14 @@ class ModUtil
         }
 
         $em = ServiceUtil::get('doctrine.entitymanager');
-        $entities = $em->getRepository('Zikula\Core\Doctrine\Entity\ExtensionVarEntity')->findBy(array('modname' => $modname, 'name' => $name));
+        $entities = $em->getRepository('Zikula\ExtensionsModule\Entity\ExtensionVarEntity')->findBy(array('modname' => $modname, 'name' => $name));
         if (count($entities) > 0) {
             foreach ($entities as $entity) {
                 // possible duplicates exist. update all (refs #2385)
                 $entity->setValue($value);
             }
         } else {
-            $entity = new \Zikula\Core\Doctrine\Entity\ExtensionVarEntity();
+            $entity = new \Zikula\ExtensionsModule\Entity\ExtensionVarEntity();
             $entity->setModname($modname);
             $entity->setName($name);
             $entity->setValue($value);
@@ -296,6 +306,9 @@ class ModUtil
 
     /**
      * The setVars method sets multiple module variables.
+     * @deprecated at Core-1.4.1
+     * @see \Zikula\ExtensionsModule\Api\VariableApi::setAll()
+     * @see service zikula_extensions_module.api.variable
      *
      * @param string $modname The name of the module.
      * @param array  $vars    An associative array of varnames/varvalues.
@@ -314,6 +327,10 @@ class ModUtil
 
     /**
      * The delVar method deletes a module variable.
+     * @deprecated at Core-1.4.1
+     * @see \Zikula\ExtensionsModule\Api\VariableApi::del()
+     * @see \Zikula\ExtensionsModule\Api\VariableApi::delAll()
+     * @see service zikula_extensions_module.api.variable
      *
      * Delete a module variables. If the optional name parameter is not supplied all variables
      * for the module 'modname' are deleted.
@@ -359,7 +376,7 @@ class ModUtil
         // else just delete this specific variable
         /** @var $qb \Doctrine\ORM\QueryBuilder */
         $qb = $em->createQueryBuilder()
-                 ->delete('Zikula\Core\Doctrine\Entity\ExtensionVarEntity', 'v')
+                 ->delete('Zikula\ExtensionsModule\Entity\ExtensionVarEntity', 'v')
                  ->where('v.modname = :modname')
                  ->setParameter('modname', $modname);
 
--- a/src/system/ExtensionsModule/Api/VariableApi.php
+++ b/src/system/ExtensionsModule/Api/VariableApi.php
@@ -1,0 +1,265 @@
+<?php
+/**
+ * Copyright 2015 Zikula Foundation
+ *
+ * This work is contributed to the Zikula Foundation under one or more
+ * Contributor Agreements and licensed to You under the following license:
+ *
+ * @license GNU/LGPLv3 (or at your option, any later version).
+ * @package Zikula
+ *
+ * Please see the NOTICE file distributed with this source code for further
+ * information regarding copyright and licensing.
+ */
+
+namespace Zikula\ExtensionsModule\Api;
+
+use Symfony\Component\HttpKernel\KernelInterface;
+use Zikula\ExtensionsModule\Entity\ExtensionVarEntity;
+use Zikula\ExtensionsModule\Entity\RepositoryInterface\ExtensionVarRepositoryInterface;
+
+/**
+ * Class VariableApi
+ * @package Zikula\ExtensionsModule\Api
+ *
+ * This class manages the storage and retrieval of extension variables and is the intended replacement
+ * for ModUtil::* methods (getVar, setVar, etc) as well as similar functionality in System:: and ThemeUtil::
+ */
+class VariableApi
+{
+    const CONFIG = 'ZConfig';
+
+    private $isInitialized = false;
+    /**
+     * @var ExtensionVarRepositoryInterface
+     */
+    private $repository;
+    /**
+     * @var KernelInterface
+     */
+    private $kernel;
+
+    /**
+     * @var array
+     */
+    private $variables;
+
+    /**
+     * ExtensionVar constructor.
+     * @param ExtensionVarRepositoryInterface $repository
+     * @param KernelInterface $kernel
+     */
+    public function __construct(ExtensionVarRepositoryInterface $repository, KernelInterface $kernel)
+    {
+        $this->repository = $repository;
+        $this->kernel = $kernel;
+    }
+
+    /**
+     * Loads extension vars for all extensions to reduce sql statements.
+     *
+     * @return void
+     */
+    private function initialize()
+    {
+        // The empty arrays for handlers and settings are required to prevent messages with E_ALL error reporting
+        $this->variables = [
+            \EventUtil::HANDLERS => [],
+            \ServiceUtil::HANDLERS => [],
+            'ZikulaSettingsModule' => [],
+        ];
+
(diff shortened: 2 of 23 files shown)


Guite:
I like your proposal.

:+1: for placing this one within the Extensions module.
We should locate each one in the corresponding module (theme -> theme, user -> user, and so on).

About the naming: how about `zikula_core.helpers.extensions.var` or even `zikula_core.extensions.helpers.var` (separating module name `extensions` from actual purpose `var`)?

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+    private $entityManager;
+    /**
+     * @var KernelInterface
+     */
+    private $kernel;
+
+    /**
+     * @var \ArrayObject
+     * @todo consider \Symfony\Component\DependencyInjection\ParameterBag\ParameterBag
+     * @todo consider \Symfony\Component\HttpFoundation\ParameterBag
```
cmfcmf: Why is this an ArrayObject in the first place? I don't see why a simple array won't work.

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+    /**
+     * ExtensionVar constructor.
+     * @param EntityManagerInterface $entityManager
+     * @param KernelInterface $kernel
+     */
+    public function __construct(EntityManagerInterface $entityManager, KernelInterface $kernel)
+    {
+        $this->entityManager = $entityManager;
+        $this->kernel = $kernel;
+        $this->initExtensionVars();
```
cmfcmf: I don't like this. IMHO the constructor shouldn't do heavy things like loading vars from the DB. The better solution would be to add an $initialized flag and load the vars in all the methods below.

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+        $this->variables = new \ArrayObject(array(
+            \EventUtil::HANDLERS => array(),
+            \ServiceUtil::HANDLERS => array(),
+            'ZikulaSettingsModule' => array(),
+        ));
+
+        // don't init vars during the installer or upgrader
+        if (!$force && \System::isInstalling()) {
+            return;
+        }
```
cmfcmf: This should just go away if possible.

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+
+        // Load all variables into the variables property.
+        /** @var ExtensionVarEntity[] $vars */
+        $vars = $this->entityManager->getRepository('Zikula\Core\Doctrine\Entity\ExtensionVarEntity')->findAll();
+        foreach ($vars as $var) {
+            if (!array_key_exists($var->getModname(), $this->variables)) {
+                $this->variables[$var->getModname()] = array();
+            }
+            if (array_key_exists($var->getName(), $GLOBALS['ZConfig']['System'])) {
```
cmfcmf: Also this - globals are legacy.

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+     * @param string $extensionName The name of the extension.
+     * @param string $variableName The name of the variable.
+     *
+     * @return boolean True if the variable exists in the database, false if not.
+     */
+    public function has($extensionName, $variableName)
+    {
+        // define input, all numbers and booleans to strings
+        $extensionName = isset($extensionName) ? ((string) $extensionName) : '';
+        $variableName = isset($variableName) ? ((string) $variableName) : '';
```
cmfcmf: IMHO this can also go away and be replaced by empty -> InvalidArgumentException checks.

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+
+        // The cast to (array) is for the odd instance where $this->variables[$extensionName] is set to null--not sure if this is really needed.
+        $varExists = isset($this->variables[$extensionName]) && array_key_exists($variableName, (array)$this->variables[$extensionName]);
```
cmfcmf: I don't see why casting to array is necessary

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+
+        if (!$varExists && \System::isUpgrading()) {
```
cmfcmf: This should go away as well.

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+     * @return boolean True if successful, false otherwise.
+     */
+    public function set($extensionName, $variableName, $value = '')
+    {
+        if (empty($extensionName) || empty($variableName)) {
+            throw new \InvalidArgumentException();
+        }
+
+        $extensionName = (string) $extensionName;
+        $variableName = (string) $variableName;
```
cmfcmf: same here

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+
+        $entities = $this->entityManager->getRepository('Zikula\Core\Doctrine\Entity\ExtensionVarEntity')
+            ->findBy(array('modname' => $extensionName, 'name' => $variableName));
+        if (count($entities) > 0) {
+            foreach ($entities as $entity) {
+                // possible duplicates exist. update all (refs #2385)
+                $entity->setValue($value);
```
cmfcmf: Logging some sort of warning here seems appropriate, if count($entities) > 1

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+     *
+     * @return boolean|null True if successful, false otherwise or null if no action.
+     */
+    public function del($extensionName, $variableName = '')
+    {
+        if (empty($extensionName)) {
+            throw new \InvalidArgumentException();
+        }
+
+        $extensionName = (string) $extensionName;
```
cmfcmf: same here

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+
+        if (!isset($this->variables[$extensionName])) {
+            return null;
```
cmfcmf: IMHO this should be return true (and also the docblock changed)

cmfcmf:
I've made some comments regarding the functionality. I understand that you much likely mostly copy-pasted from ModUtil for now to clarify how everything should work and probably didn't check for legacy code yet.

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+    {
+        if (empty($extensionName)) {
+            throw new \InvalidArgumentException();
+        }
+
+        // if vars for this extension not available then get them
+        if (!array_key_exists($extensionName, $this->variables)) {
+            // A query out to the database should only be needed if the system is upgrading. Use the installing flag to determine this.
+            if (\System::isUpgrading()) {
+                $this->initExtensionVars(true);
```
cmfcmf: Also should go away

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+            } else {
+                // Prevent a re-query for the same extension in the future, where the extension does not define any extension variables.
+                $this->variables[$extensionName] = array();
+            }
+        }
+
+        // if variable name not passed, then return every variable for the specified extension as an associative array.
+        if (empty($variableName) && array_key_exists($extensionName, $this->variables)) {
```
cmfcmf: Maybe extract this to a *getAll* method and disallow empty variable names for the get method.

src/lib/Zikula/Core/Helper/ExtensionVar.php
craigh: copying previous implementation in ModUtil.

src/lib/Zikula/Core/Helper/ExtensionVar.php
craigh: they are casting to strings. I believe this is important.

src/lib/Zikula/Core/Helper/ExtensionVar.php
craigh: again casting

src/lib/Zikula/Core/Helper/ExtensionVar.php
cmfcmf: I know they are casting to strings - but why? If

craigh: I guess to ensure other types don't get passed to the DB. you cannot enforce `(string)` args in a function until php 7 so this is the best way for now. to rely on the end developer doing it right is probably not wise.

src/lib/Zikula/Core/Helper/ExtensionVar.php
```diff
@@ -0,0 +1,283 @@
...
+    private $variables;
+
+    /**
+     * ExtensionVar constructor.
+     * @param EntityManagerInterface $entityManager
+     * @param KernelInterface $kernel
+     */
+    public function __construct(EntityManagerInterface $entityManager, KernelInterface $kernel)
+    {
+        $this->entityManager = $entityManager;
```
cmfcmf: It would be better to inject the EntityVarRespository instead. It simplifies testing the code. http://php-and-symfony.matthiasnoback.nl/2014/05/inject-a-repository-instead-of-an-entity-manager/ (also see the last paragraph)

src/system/ExtensionsModule/Api/VariableApi.php
```diff
@@ -0,0 +1,256 @@
...
+     * Checks to see if an extension variable is set.
+     *
+     * @param string $extensionName The name of the extension.
+     * @param string $variableName The name of the variable.
+     *
+     * @return boolean True if the variable exists in the database, false if not.
+     */
+    public function has($extensionName, $variableName)
+    {
+        if (!is_string($extensionName) || !is_string($variableName)) {
```
cmfcmf: IMHO this should also check for empty.

src/system/ExtensionsModule/Entity/Repository/ExtensionVarRepository.php
```diff
@@ -0,0 +1,58 @@
...
+use Zikula\ExtensionsModule\Entity\RepositoryInterface\ExtensionVarRepositoryInterface;
+
+class ExtensionVarRepository extends EntityRepository implements ExtensionVarRepositoryInterface
+{
+    public function remove($entity)
+    {
+        $this->_em->remove($entity);
+    }
+
+    public function save($entity)
```
cmfcmf: Maybe call it `persistAndFlush` instead, because it clarifies what it actually does. Otherwise one might think that calling `remove` will flush as well.

cmfcmf:
:+1: Now tests are the only thing missing.

src/system/ExtensionsModule/Api/VariableApi.php
```diff
@@ -0,0 +1,255 @@
...
+     * @param string $extensionName The name of the extension.
+     * @param array $vars An associative array of varnames/varvalues.
+     *
+     * @return boolean True if successful, false otherwise.
+     */
+    public function setAll($extensionName, array $vars)
+    {
+        $ok = true;
+        foreach ($vars as $var => $value) {
+            $ok = $ok && $this->set($extensionName, $var, $value);
```
craigh: @cmfcmf I think this logic is a remnant of the ArrayObject. wouldn't it be sufficient to just do:

`$this->variables[$extensionName] = $vars;` and also update the DB?

cmfcmf: This does change the logic: Your approach would also delete all other variables of the extension. The current approach does not touch variables not present in the $vars array.

craigh: good point. I guess I leave it.

craigh:
@Guite @cmfcmf @shefik @Kaik @phaidon @espaan 

you :+1: or :-1: or comments on this PR are appreciated. Merging soon.

Guite:
:+1: 

Using this is done with `zikula_extensions_module.api.variable` service?

craigh:
> Using this is done with zikula_extensions_module.api.variable service?

for now, that is the only way. I will add some shortcuts to the Abstract Controller though.

I also just realized that maybe I should add tests for the Repo. @cmfcmf agree?

src/lib/util/ModUtil.php
```diff
@@ -296,6 +306,9 @@ public static function setVar($modname, $name, $value = '')
 
     /**
      * The setVars method sets multiple module variables.
+     * @deprecated at Core-1.4.1
+     * @see \Zikula\ExtensionsModule\Api\VariableApi::setAll()
```
cmfcmf: Again - unsure whether `@see` breaks annotation readers.

src/system/ExtensionsModule/Api/VariableApi.php
```diff
@@ -0,0 +1,265 @@
...
+            $entity->setModname($extensionName);
+            $entity->setName($variableName);
+        } else {
+            $entity = $entities[0];
+        }
+        $entity->setValue($value);
+        $this->repository->persistAndFlush($entity);
+        $this->variables[$extensionName][$variableName] = $value;
+
+        return true;
```
cmfcmf: One might argue that nothing should be returned, because:
- it returns true all the time anyways
- if a developer cleanly uses the set function, they are forced to check for the return type and if it's false do... what?

src/lib/util/ModUtil.php
craigh: this is the proper use of `@see`

http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.see.pkg.html

cmfcmf: Whoops, you are right. Apparently, `@use` is the evil one: https://github.com/zikula/core/pull/1517/files#diff-aca64025345ee21fbaa0d4436006a52fL266

PR #2627, opened 2015-10-13
Refactor to eliminate the GroupMembershipEntity. 

craigh:
This should have been done when converting from legacy systems to Doctrine. The Entity doesn't exist as an entity in and of itself, but rather as a join table which can be accomodated within the definitions of the UserEntity and GroupEntity.

**Please extensively test all User/Group functionality with this PR.**
ping @cmfcmf @Guite @shefik @Kaik @phaidon @espaan 

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | yes, technically.
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | yes


src/system/GroupsModule/Entity/GroupEntity.php
```diff
@@ -313,4 +324,32 @@ public function setUidmaster($uidmaster)
...
+     * @return ArrayCollection
+     */
+    public function getUsers()
+    {
+        return $this->users;
+    }
+
+    /**
+     * GroupEntity is the 'Inverse side'
+     * @see http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/association-mapping.html#owning-and-inverse-side-on-a-manytomany-association
```
cmfcmf: Not sure whether `@see` might break annotation readers.

PR #2630, opened 2015-10-15
[WIP] create PermissionApi

craigh:
unfinished. comments welcome

### DIFF
--- a/src/system/PermissionsModule/Api/PermissionApi.php
+++ b/src/system/PermissionsModule/Api/PermissionApi.php
@@ -1,0 +1,354 @@
+<?php
+/**
+ * Copyright Zikula Foundation 2015 - Zikula Application Framework
+ *
+ * This work is contributed to the Zikula Foundation under one or more
+ * Contributor Agreements and licensed to You under the following license:
+ *
+ * @license GNU/LGPLv3 (or at your option, any later version).
+ *
+ * Please see the NOTICE file distributed with this source code for further
+ * information regarding copyright and licensing.
+ */
+
+namespace Zikula\PermissionsModule\Api;
+
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\Translation\TranslatorInterface;
+use Zikula\PermissionsModule\Entity\RepositoryInterface\PermissionRepositoryInterface;
+use Zikula\UsersModule\Entity\RepositoryInterface\UserRepositoryInterface;
+use Zikula\UsersModule\Entity\UserEntity;
+
+/**
+ * Class PermissionApi
+ * @package Zikula\PermissionsModule\Api
+ *
+ * This class is used to determine whether a user has rights (or permissions) to a given component. Rights are granted
+ * or denied from the Permissions module User Interface. Components/Extensions must declare their Permission structure in
+ * their `composer.json` file.
+ */
+class PermissionApi
+{
+    /**
+     * 'all users', includes unregistered users
+     */
+    const ALL_USERS = -1;
+    /**
+     * unregistered user
+     */
+    const UNREGISTERED_USER = 0;
+    /**
+     * 'all groups', includes unregistered users
+     */
+    const ALL_GROUPS = -1;
+    /**
+     * unregistered users.
+     */
+    const UNREGISTERED_USER_GROUP = 0;
+
+    /**
+     * @var PermissionRepositoryInterface
+     */
+    private $permRepository;
+
+    /**
+     * @var UserRepositoryInterface
+     */
+    private $userRepository;
+
+    /**
+     * @var SessionInterface
+     */
+    private $session;
+
+    /**
+     * @var \Zikula\Common\Translator\Translator
+     */
+    private $translator;
+
+    /**
+     * holds the cache of group Perms by User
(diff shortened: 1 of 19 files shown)


src/system/UsersModule/Entity/Repository/UserRepository.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Zikula\UsersModule\Entity\Repository;
+
+use Doctrine\ORM\EntityRepository;
+use Zikula\UsersModule\Entity\UserEntity;
+use Zikula\UsersModule\Entity\RepositoryInterface\UserRepositoryInterface;
+
+class UserRepository extends EntityRepository implements UserRepositoryInterface
+{
+    public function findByUids($uids)
```
cmfcmf: Matthias Noback even suggested to leave out the "find" part of the method names in repositories, because finding things is tightly coupled to the idea of doctrine searching the database.

src/system/PermissionsModule/Api/PermissionApi.php
```diff
@@ -0,0 +1,323 @@
...
+    public function hasPermission($component = null, $instance = null, $level = ACCESS_NONE, $user = null)
+    {
+        if (!is_numeric($level)) {
+            throw new \InvalidArgumentException('Invalid security level');
+        }
+        if (isset($user) and !is_numeric($user)) {
+            throw new \InvalidArgumentException('User argument must be an integer.');
+        }
+        if (!isset($user)) {
+            $user = $this->session->get('uid', false);
```
cmfcmf: There might not be a session (for example when using the CLI). It should be checked what symfony's security component does in this case (e.g. Assume admin permission or something).

src/system/PermissionsModule/Api/PermissionApi.php
```diff
@@ -0,0 +1,323 @@
...
+                if ($perm['level'] > $level) {
+                    $level = $perm['level'];
+                }
+            }
+
+            return $level;
+        }
+
+        // Test if user has ANY access to given component, without determining exact instance
+        if ($instance == 'ANY') {
```
cmfcmf: Maybe make it a constant

src/system/PermissionsModule/Api/PermissionApi.php
```diff
@@ -0,0 +1,323 @@
...
+        $accessnames = [
+            ACCESS_NONE => $this->translator->__('No access'),
+            ACCESS_OVERVIEW => $this->translator->__('Overview access'),
+            ACCESS_READ => $this->translator->__('Read access'),
+            ACCESS_COMMENT => $this->translator->__('Comment access'),
+            ACCESS_MODERATE => $this->translator->__('Moderate access'),
+            ACCESS_EDIT => $this->translator->__('Edit access'),
+            ACCESS_ADD => $this->translator->__('Add access'),
+            ACCESS_DELETE => $this->translator->__('Delete access'),
+            ACCESS_ADMIN => $this->translator->__('Admin access'),
```
cmfcmf: Missing ACCESS_INVALID? When is it even used..

src/system/UsersModule/Entity/Repository/UserRepository.php
craigh: I'm not sure I understand what you mean. Do you mean simply renaming the method? To what would you suggest? I'm not sure I would agree, because no matter the repository, the method indeed is *find*ing users by a `uid` array.

src/system/PermissionsModule/Api/PermissionApi.php
craigh: I can't find anything...

src/system/PermissionsModule/Api/PermissionApi.php
craigh: this method is used once or twice in the core and once in News that I know of. (If that was your question)

cmfcmf: Sorry, I was unclear. I meant "when is ACCESS_INVALID" even used.

src/system/UsersModule/Entity/Repository/UserRepository.php
cmfcmf: Yes, simply renaming to `byUids`. If Doctrine would not have been used, the entities might not be "found" but maybe "generated" or "streamed" or "read". "Finding" is a bit more Doctrine and database specific.

Don't get me wrong here. I'm perfectly fine with the current naming. Chances that Doctrine isn't used are rather small anyways. Just wanted to present the option to rename the method.

src/system/PermissionsModule/Api/PermissionApi.php
cmfcmf: Hmm, I will look as well. Questions to answer for Symfony:
- Is there a request object if the CLI is used?
- Is there a session object if the CLI is used?
- [How does this behave if called via CLI](http://symfony.com/doc/current/book/security.html#securing-controllers-and-other-code)?

craigh:
@Guite @Kaik @phaidon @cmfcmf I've got this mostly done. There are a few things that maybe should get finished before merging. your review would be appreciated. You can also clone and run the tests if you like.

PR #2633, opened 2015-11-02
update to MOST 2472ad3

craigh:
update to MOST 2472ad3

@cmfcmf I'm not sure I did this right. It adds many `.generated` files which I suspect are not needed. So, I can update the PR if needed. But I also notice some other things that made the process challenging.

@Guite - one thing you can see here that is wrong is the changing of paths from `system/RoutesModule/foo` to `system/ZikulaRoutesModule/foo` or `system/zikulaRoutes/foo` (note case). Maybe this should be corrected in the generator?

Also - another thing that was difficult for me in the diff was that the line-endings were different and so most files appeared to be different but in reality were not. I'm not sure if this is my fault (MacOS) or how the files are generated (windows line endings?). Maybe this can be improved.

Anyway - this is obviously incomplete and so it requires adjustment, but you can simply pull the branch and change it if you want, or let me know what needs to be done.


Guite:
The path change is covered by https://github.com/Guite/MostGenerator/issues/686

cmfcmf:
Thank you Craig!

craigh:
I am learning about MOST as I look at this.

I think there is a strong case for maintaining unmodified copies of the `.generated` files in the module. Mainly, this would allow for easier updates - e.g. if the `.generated` is changed (which would be seen in the diff) then those same changes would need to be made in the customized classes/methods.

thoughts?

also - I do not understand why a second `system/RoutesModule/Resources/docs/model/Routes2.mostapp` has been generated. the first (`/Applications/MAMP/htdocs/core.git/src/system/RoutesModule/Resources/docs/model/Routes.mostapp`) appears to be exactly the same. Is there supposed to be some difference here?

cmfcmf:
> I am learning about MOST as I look at this.
> 
> I think there is a strong case for maintaining unmodified copies of the .generated files in the module. Mainly, this would allow for easier updates - e.g. if the .generated is changed (which would be seen in the diff) then those same changes would need to be made in the customized classes/methods.

Yes. I don't know of potential disadvantages.
> 

> also - I do not understand why a second system/RoutesModule/Resources/docs/model/Routes2.mostapp has been generated. the first (/Applications/MAMP/htdocs/core.git/src/system/RoutesModule/Resources/docs/model/Routes.mostapp) appears to be exactly the same. Is there supposed to be some difference here?

A guess: Did you call the model Routes2 at the webgen?

craigh:
> A guess: Did you call the model Routes2 at the webgen?

I did not.

screen shot 2015-11-03 at 2 06 19 pm

craigh:
now I see the work that remains - the current `Util` classes in the module must be refactored to the new `Helper` classes that were added! I see @Guite 's wizardly ways... I guess there is more work here than I originally imagined, but this is probably a good thing.

craigh:
obviously this will require some serious changes through the module because these are used as services. My question though is, does the **model** require changes first? I ask because I see at least one [mentioned in the model file](https://github.com/zikula/core/blob/MOST-UPDATE-Nov2/src/system/RoutesModule/Resources/docs/model/Routes2.mostapp#L49). @Guite what am I supposed to do here?

craigh:
@cmfcmf I think I must leave the comparison of the templates to you. I'm not exactly sure what I am looking for.

also - leaving the `.generated` template files is no problem (and IMO, a good idea to track changes) but the `.generated` class files may not work because the classnames are the same in both files ([normal](https://github.com/zikula/core/blob/MOST-UPDATE-Nov2/src/system/RoutesModule/Controller/AjaxController.php#L13), [generated](https://github.com/zikula/core/blob/MOST-UPDATE-Nov2/src/system/RoutesModule/Controller/AjaxController.generated.php#L13)) and at least my IDE was being fooled on the location. maybe PHP will also have issues. I haven't tested. Since the filename is different maybe the autoloader will ignore it. maybe.

craigh:
also - stuff like [this](https://github.com/zikula/core/pull/2633/files#diff-34bbc153c86a999b08a70c5fd543b981R1) still has to be corrected.

cmfcmf:
Regarding the .generated files: could be fixed if they were called test.php.generated instead of test.generated.php.

Guite:
Better would be removing them in CI.

craigh:
> Better would be removing them in CI.

removing them in CI is not a solution if the local dev environment cannot operate properly. As I said, I have not tested it at all, but I don't think that is a proper solution either. How do you feel about a different name-scheme?

craigh:
I've done some limited usage testing locally and the `.generated` files don't seem to be conflicting with anything and so it appears there is no harm in leaving them here.

Should I revert the file perms? Or do we not care?

cmfcmf:
> Should I revert the file perms? Or do we not care?

I think we care.

PR #2644, opened 2015-11-21
fix route translation

### LINKED ISSUE #2642: admin panel with german translation can not be opened
after finalizing the installation I am guided to http://koester123.de/admin/panel

In the past I was request to log in. But this time I did get a message by firefox:

````
Fehler: Umleitungsfehler

Die aufgerufene Website leitet die Anfrage so um, dass sie nie beendet werden kann.

    Dieses Problem kann manchmal auftreten, wenn Cookies deaktiviert oder abgelehnt werden.
````

If I go back to homepage (deleting /admin/panel) I see the homepage and con log in. If I than try to go to admin site I am back into the same loop.

I am using ci build 113 with some German translation files (which are working in the past pretty well)

craigh:
Remove all multilingual and prefix options from controller methods responsible to change the locale. Set default locale to 'en' where possible. Ensure module metadata 'ur' is translated properly for each locale's route. fixes #2642.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | n/a
| Fixed tickets     | #2642
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | not req.

### DIFF
--- a/src/app/config/config.yml
+++ b/src/app/config/config.yml
@@ -4,7 +4,7 @@ imports:
 
 framework:
     #esi:             ~
-    translator:      { fallbacks: %locale% }
+    translator:      { fallbacks: en }
     secret:          %secret%
     router:
         resource: "%kernel.root_dir%/config/routing.yml"
--- a/src/lib/Zikula/Bundle/CoreBundle/Bundle/MetaData.php
+++ b/src/lib/Zikula/Bundle/CoreBundle/Bundle/MetaData.php
@@ -124,10 +124,10 @@ class MetaData
         return $this->__(/** @Ignore */$this->displayName);
     }
 
-    public function getUrl()
+    public function getUrl($translated = true)
     {
         $this->confirmTranslator();
-        return $this->__(/** @Ignore */$this->url);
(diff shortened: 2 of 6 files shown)


craigh:
I believe this is ready for merging. Objections? @cmfcmf @Kaik @Guite @rallek

craigh:
It occurs to me that I don't know how this will work with third party modules - especially legacy type modules like News. If you can test that, I would appreciate it.

craigh:
also - please note how I changed the usage of the commit message template. I have added the standard table to the first message in the PR, but I would like to stop doing that and add it right to the commit like I did [here](https://github.com/zikula/core/commit/39573656b14398dd2a59e6f75fa1ad8ed445f280).

This is how symfony does it also. I like this because it will be available in the git commit log which the ticket is not. Let me know what you think about this idea.

craigh:
> It occurs to me that I don't know how this will work with third party modules

I just did some quick tests with News and Pages and they seemed to work fine. They don't have translations though, so everything is not translated, but otherwise fine.

PR #2645, opened 2015-11-22
Enable language to be re selected in installation.

Kaik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | Yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     |
| Refs tickets      | #2642
| License           | MIT
| Doc PR            | 
| Changelog updated | no

This setting should be to true so language chooser page will show up.

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Stage/LocaleStage.php
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Stage/LocaleStage.php
@@ -71,7 +71,7 @@ class LocaleStage implements StageInterface, FormHandlerInterface, InjectContain
             if (($locale !== false) && (in_array($locale, $installedLanguages))) {
                 $this->writeParams(array('locale' => $locale));
 
-                return false;
+                return true;
             } else {
 
                 return true;


src/lib/Zikula/Bundle/CoreInstallerBundle/Stage/LocaleStage.php
```diff
@@ -71,7 +71,7 @@ public function isNecessary()
             if (($locale !== false) && (in_array($locale, $installedLanguages))) {
                 $this->writeParams(array('locale' => $locale));
 
-                return false;
+                return true;
```
craigh: I'm not sure this is correct. The method should return `false` is the method is **not** necessary. The code immediately before this change set the locale based on browser detection and so it would seem that the stage is **not** necessary because the task of setting the locale is **complete**. returning `true` will display the locale selector, which seems not necessary if it has been detected from the browser... ?

Kaik: Well this is installer and default locale for website should not be chosen without "option" to change it specially if it is detected from browser. For example let say you have de in browser but site still should be in en... you can't go back Language chooser is not displayed. This setting should be true for any case where there is more than one language installed.

craigh:
> default locale for website should not be chosen without "option" to change it specially if it is detected from browser.

If this is the desired behavior, then the whole block of code that checks for browser language should be removed. Simply returning `true` isn't the correct fix. @rallek @Guite @Kaik @cmfcmf what is your preference for behavior?

Guite:
I'd prefer a possibility to change the language manually, but having the browser language selected by default.

craigh:
that's actually harder than you might think.

rallek:
We are talking about installation, right? Than the question have to start earlier. What kind of downloads we want to offer? Currently I have to download at ci the core and to look where I can find my german translation. During this way I have to use a lot of English based pages. If this should remain for 1.4.x we do not need a translation for installation at all.

But this will limit our community. If we can offer a complete package by our German community I would expect a German dialog for installation. If this is done by browser language detection or if it is "hard coded" in the complete package it is not important for me.

I do not expect a package with more than two languages for installation. It might be different for the site itself. There might be some more languages.

Another situation would be also possible. Let us think about a multilingual installer with n translations by default. Than I have to choose my language at the beginning. This might preselected by the browser. But the core do not have all these languages included already. There will be the possibility to install via a new function inside the core additional languages e.g. from a central language repository. In this case the installation language should be downloaded and installed automatically. But I think this new situation is not possible in short time. 

At the end my summary: We do not have an idea how to manage the language packs at all. Without this we do not need a translated installer for the moment. Let us focus on a working translation for the core itself.

Kaik:
I think it remembers it you can click on Select language pill and there you have it.

Kaik:
@craigh That block is important it sets locale to the browser one or chosen before I think...

Kaik:
@rallek I'm working on sort of language management but that will work after install. The whole thing is to distinguish languages from translations - translation for example there can be 4 language's on site and each of them will have different status of translations for each module so translation is the property of language and those translations are catalogued by so called "domain" for example zikula routes validators - with new system we can trace file presence and its content if its "not empty" untranslated routes can eliminate language to be set as "routing language"... long story :)

craigh:
> That block is important it sets locale to the browser one or chosen before I think...

The code in the if/else block sets the `locale` param (in `custom_parameters.yml`) to the browser-detected locale. but it won't set that value to the 'selected' value in the form. at least not yet.

part of the complication is that the for is used in both installation and upgrade. For upgrade, it has to use the previously selected/installed locale...

cmfcmf:
> I'd prefer a possibility to change the language manually, but having the browser language selected by default.

:+1:

Kaik:
Ok so show for installer default browser (I do not know about previous selected should it be current locale?) and for upgrade from previous install but that may generate errors when no translations will be installed for previous language...

craigh:
closed in favor of #2796

PR #2683, opened 2015-12-16
Fix stylesheet

cmfcmf:


craigh:
I think I might prefer a `file_exists` check in the AbstractModule

craigh:
merge = bad. rebase = good.

cmfcmf:
Thank you :P

Craig Heydenburg <notifications@github.com> schrieb am Do., 17. Dez. 2015
um 15:05 Uhr:

cmfcmf:
> I think I might prefer a file_exists check in the AbstractModule

I was considering the same thing, but went with the exception because it already does the check in `locateResource`.

Issue #2663, opened 2015-12-05
filter incompatibility with Symfony 2.8


craigh:
at the route `/routes/admin` I get this error:

> [1] Symfony\Component\Debug\Exception\ContextErrorException: User Error: Passing the $deep boolean as 3rd argument to the Symfony\Component\HttpFoundation\ParameterBag::filter method is deprecated since version 2.8 and will be removed in 3.0. Remove it altogether as the $deep argument will be removed in 3.0.
    at n/a
        in /Applications/MAMP/htdocs/core.git/src/vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/ParameterBag.php line 271

>    at Symfony\Component\Debug\ErrorHandler->handleError('256', 'Passing the $deep boolean as 3rd argument to the Symfony\Component\HttpFoundation\ParameterBag::filter method is deprecated since version 2.8 and will be removed in 3.0. Remove it altogether as the $deep argument will be removed in 3.0.', '/Applications/MAMP/htdocs/core.git/src/vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/ParameterBag.php', '271', array('key' => 'ot', 'default' => 'route', 'filter' => false, 'options' => '513', 'deep' => array(), 'filters' => array('1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1'), 'tmp' => 'callback'))
        in  line 

>    at trigger_error('Passing the $deep boolean as 3rd argument to the Symfony\Component\HttpFoundation\ParameterBag::filter method is deprecated since version 2.8 and will be removed in 3.0. Remove it altogether as the $deep argument will be removed in 3.0.', '256')
        in /Applications/MAMP/htdocs/core.git/src/vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/ParameterBag.php line 271

>    at Symfony\Component\HttpFoundation\ParameterBag->filter('ot', 'route', false, '513', array())
        in /Applications/MAMP/htdocs/core.git/src/lib/legacy/Zikula/Bag/ParameterBag.php line 54

>    at Zikula_Bag_ParameterBag->filter('ot', 'route', false, '513')
        in /Applications/MAMP/htdocs/core.git/src/system/RoutesModule/Controller/Base/AdminController.php line 64

>    at Zikula\RoutesModule\Controller\Base\AdminController->indexAction(object(Zikula_Request_Http))
[...]

PR #2686, opened 2015-12-17
Correct usage of `$deep` parameter for Symfony 2.8. Refs #2663

craigh:


craigh:
@Guite @cmfcmf @Kaik thoughts on this? Christian mentioned he thought it might be a BC break for us. I'm not sure that it is because I don't think Symfony will do anything with the `$deep` parameter anyways. So pointless to send it on.

Guite:
As long as you do not change offset values for the further arguments (filter and options) existing usages with `deep=false` should still work.

There seems a BC break for usages with `deep=true` though. At least I did not use this anywhere.

Issue #2672, opened 2015-12-13
JS routes in multilingual setup not working properly.


craigh:
Page not found. :frowning: 

> [1] Symfony\Component\HttpKernel\Exception\NotFoundHttpException: Page not found.
    at n/a
        in /Applications/MAMP/htdocs/core.git/src/lib/Zikula/Bundle/CoreBundle/EventListener/LegacyRouteListener.php line 95

>    at Zikula\Bundle\CoreBundle\EventListener\LegacyRouteListener->onKernelRequest(object(GetResponseEvent), 'kernel.request', object(TraceableEventDispatcher))
        in  line 

>    at call_user_func(array(object(LegacyRouteListener), 'onKernelRequest'), object(GetResponseEvent), 'kernel.request', object(TraceableEventDispatcher))
        in /Applications/MAMP/htdocs/core.git/src/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/Debug/WrappedListener.php line 61

>    at Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke(object(GetResponseEvent), 'kernel.request', object(Zikula_EventManager))
        in  line 

>    at call_user_func(object(WrappedListener), object(GetResponseEvent), 'kernel.request', object(Zikula_EventManager))
        in /Applications/MAMP/htdocs/core.git/src/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php line 181

>    at Symfony\Component\EventDispatcher\EventDispatcher->doDispatch(array(object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener)), 'kernel.request', object(GetResponseEvent))
        in /Applications/MAMP/htdocs/core.git/src/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php line 46

>    at Symfony\Component\EventDispatcher\EventDispatcher->dispatch('kernel.request', object(GetResponseEvent), null)
        in /Applications/MAMP/htdocs/core.git/src/lib/legacy/Zikula/EventManager.php line 99

[...]

PR #2711, opened 2015-12-26
Make cookie httponly = false.

cmfcmf:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | ---
| Fixed tickets     | ---
| Refs tickets      | ---
| License           | MIT
| Doc PR            | ---
| Changelog updated | no

### DIFF
--- a/src/app/config/config.yml
+++ b/src/app/config/config.yml
@@ -26,6 +26,7 @@ framework:
         name: %zikula.session.name%
         handler_id: session.handler.legacy
         storage_id: session.storage.legacy
+        cookie_httponly: false
     fragments:       ~
 
 # Twig Configuration
--- a/src/lib/legacy/Zikula/Controller/AbstractAjax.php
+++ b/src/lib/legacy/Zikula/Controller/AbstractAjax.php
@@ -29,7 +29,7 @@ abstract class Zikula_Controller_AbstractAjax extends Zikula_AbstractController
     }
 
     /**
-     * Check the CSFR token.
+     * Check the CSRF token.
      *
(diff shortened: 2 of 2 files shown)


cmfcmf:
I would be glad if @craigh could write a test for this? :pray:

cmfcmf:
> I would be glad if @craigh could write a test for this? :pray: 
Whoops, wrong ticket.

src/app/config/config.yml
```diff
@@ -26,6 +26,7 @@ framework:
         name: %zikula.session.name%
         handler_id: session.handler.legacy
         storage_id: session.storage.legacy
+        cookie_httponly: false
```
cmfcmf: If this is set to false, the session id cannot be retrieved in JavaScript via `document.cookie`. This is currently required however to make our AJAX requests work.

craigh: I don't understand. you are setting it to false, but saying that it doesn't work if set to false?

craigh:
is this related at all to https://github.com/zikula/core/issues/2672 ?

cmfcmf:
> is this related at all to #2672 ?

I don't know. I'm not working in a multilingual setup at the moment. With this change `$this->checkAjaxToken()` works again.

craigh:
> `$this->checkAjaxToken()` works again

:+1: :tada:

src/app/config/config.yml
cmfcmf: I mixed up true and false. Updated the comment.

PR #2718, opened 2016-01-06
Fixed description

GrahamCampbell:


GrahamCampbell:
The remaining StyleCI patch should now be safe to apply.

GrahamCampbell:
It can send you a pull request if you click the "Send fix PR" button on the most recent analysis. ;)

craigh:
yes. thanks I'll take a look.

craigh:
@GrahamCampbell - it is removing default values of method arguments for some reason. Is this a feature or a bug? Any way to disable?

GrahamCampbell:
> it is removing default values of method arguments for some reason. Is this a feature or a bug? Any way to disable?

It's because they are impossible default values because they are followed by a value without a default. Do you see?

GrahamCampbell:
eg, the following default value is impossible:

```php
function test($foo = null, $bar) {}
```

craigh:
thanks. I see now. A lot of old code... grrr.

Issue #2722, opened 2016-01-07
Wrong grid class used?


Guite:
In https://github.com/zikula/core/blob/1.4/src/lib/Zikula/Bundle/FormExtensionBundle/Resources/views/Form/bootstrap_3_zikula_admin_layout.html.twig there are `col-lg-*` classes used. Shouldn't we prefer `col-sm-*` (or at least `col-md-*`)?

craigh:
closed by #2726

PR #2726, opened 2016-01-09
use grid classes for smaller viewport

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2722
| Refs tickets      | 
| License           | MIT
| Doc PR            | 
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/FormExtensionBundle/Resources/views/Form/bootstrap_3_zikula_admin_layout.html.twig
+++ b/src/lib/Zikula/Bundle/FormExtensionBundle/Resources/views/Form/bootstrap_3_zikula_admin_layout.html.twig
@@ -1,11 +1,11 @@
 {% extends "bootstrap_3_horizontal_layout.html.twig" %}
 
 {% block form_label_class -%}
-col-lg-3
+col-sm-3
 {%- endblock form_label_class %}
 
 {% block form_group_class -%}
-col-lg-9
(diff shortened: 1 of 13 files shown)


craigh:
@Guite I'm just looking into this. but I see you did not create a "topic" branch. this makes checking out the PR much more difficult. In the future, please create topic branches if you intend to PR.

craigh:
before
after

craigh:
I'm not sure which I prefer. I think I remember a previous discussion about this a couple years ago. may @phaidon was involved? I can't be sure, it has been a long time ago. If the field label is extra long, the current method may be better since the break would occur sooner? maybe @Kaik has an opinion or @cmfcmf or @shefik or @rallek

rallek:
for me both cases are fine. I do not have enough experience to judge this. Might be while testing the current 1.4.2dev I do get an idea of this.

Kaik:
I prefer second option - might be, a step sooner(sm), but first one is too soon I think...

craigh:
@Guite maybe `col-sm-*` would be better? (I really don't know)

craigh:
here is one of the commits I was thinking about. I guess it wasn't so long ago :wink: 
https://github.com/zikula/core/pull/2399

PR #2745, opened 2016-01-18
incremented dep version for distribution bundle, fixed #2656

### LINKED ISSUE #2656: update SensioDistributionBundle
https://github.com/sensiolabs/SensioDistributionBundle/releases

current: `"sensio/distribution-bundle": "4.0.*",`

available: 5.0.3

there seems to be no documentation or changelog regarding differences between v4 and v5.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2656
| Refs tickets      |
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "symfony/assetic-bundle": "2.6.*",
         "symfony/swiftmailer-bundle": "2.3.*",
         "symfony/monolog-bundle": "2.7.*",
-        "sensio/distribution-bundle": "4.0.*",
+        "sensio/distribution-bundle": "5.0.*",
         "sensio/framework-extra-bundle": "3.0.*",
         "elao/web-profiler-extra-bundle" : "dev-master",
         "matthiasnoback/symfony-service-definition-validator": "1.2.*",


craigh:
did you try this locally? Any changes/errors?

Guite:
Yes I tried `composer install`. No obvious changes, no errors.

craigh:
I'm trying this locally also. I've not fully tested anything yet, but I found this: https://github.com/symfony/symfony-standard/pull/890/files

which indicates changes to bootstrap.php.cache but I do not know what this file is used for. After updating to 5.0.*, I see that this file no longer contains the following:

```
namespace {
    error_reporting(error_reporting() & ~E_USER_DEPRECATED);
    $loader = require_once __DIR__.'/./autoload.php';
}
```
(this is at the very top of the file)

and

```
namespace { return $loader; }
```
(this is at the very end of the file)

so *if* this file is used for something, then it appears to no longer load the autoloader file. But I don't know if that is a problem.

craigh:
I guess since zikula has it's [own bootstrap that requires the autoloader](https://github.com/zikula/core/blob/1.4/src/lib/bootstrap.php#L19) then maybe this isn't an issue?

PR #2747, opened 2016-01-19
added Twig support to display hook response, fixed #2600

### LINKED ISSUE #2600: Refactor DisplayHookResponse so that you don't have to use Smarty
Currently, you have to use Smarty:
- https://github.com/zikula/core/blob/1.4/src/lib/Zikula/Core/Hook/DisplayHookResponse.php#L55
- https://github.com/zikula/core/blob/1.4/src/lib/Zikula/Core/Hook/DisplayHookResponse.php#L138

This should be changed to Twig, or even better, to just "strings", so that the response isn't aware of the generation of the response.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | ?
| Fixed tickets     | #2600
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

@craigh note the topic branch :smile:

### DIFF
--- a/src/lib/Zikula/Core/Hook/DisplayHookResponse.php
+++ b/src/lib/Zikula/Core/Hook/DisplayHookResponse.php
@@ -32,9 +32,9 @@ class DisplayHookResponse
     protected $area;
 
     /**
-     * Zikula_View
+     * The rendering engine.
      *
-     * @var Zikula_View
+     * @var Zikula_View|Twig_Environment
      */
     protected $view;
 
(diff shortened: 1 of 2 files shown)


cmfcmf:
Have you actually tested this? How would I pass parameters to the template? What if I didn't want to use Twig but rather PHP?

craigh:
@cmfcmf has a point. In Smarty, the hookHandler could simply assign the variables to the View object and then pass that object along to the DisplayHookReponse. But in Twig, this isn't possible, you have to pass the variables as a separate argument in the render method. So this PR alone will not be sufficient, we need more code. 

But @cmfcmf I'm not sure it is reasonable to allow for PHP templating. The rest of zikula assumes Twig usage. It is reasonable to assume that in this case. Zikula cannot accommodate *every* option certainly.

cmfcmf:
> The rest of zikula assumes Twig usage.

Does it? Can't I return anything I want from, e.g., a Controller?

PR #2767, opened 2016-02-13
Prevent a potential timeout, if the site is using a load balancer.

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxInstallController.php
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxInstallController.php
@@ -192,6 +192,17 @@ class AjaxInstallController extends AbstractController
                 'state' => \ModUtil::STATE_INACTIVE));
             $modApi->setstate(array('id' => $mid,
                 'state' => \ModUtil::STATE_ACTIVE));
+                
+            /**
+             * Prevent a potential timeout, if the site is using a load balancer.
+             */
+            sleep(1);
+
+            if (!headers_sent()) {
+                header('Content-Type: application/json');
+            }
+
+            flush();
         }
 
         return true;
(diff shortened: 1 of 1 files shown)


craigh:
Mrs must pass all checks before merging. please correct the StyleCI issue.

craigh:
I don't really understand this well, so more information would be appreciated. But, that being said, as I understand it, this would add 14 seconds to this stage of the installation. From a Users perspective this is a really long time and I don't think that would be the best solution. Is there any way to *detect* if a load balancing system is in place? And if so, make adjustments then? This seems like a hacky solution - maybe we can address the real problem, if we can know it.

shefik:
@craigh Fixed.

shefik:
The main issue is with a host such as Rackspace, which has a load balancer on their Cloud Sites service. If there is a long AJAX execution, it does not matter if the AJAX timeout is set longer. Eventually, the load balancer will timeout, since it is not sending a response back to the browser. So, this pull request periodically sends a header, to indicate that it should not timeout the load balancer.

craigh:
@Guite thoughts?

craigh:
@shefik PR needs reset.

craigh:
@Guite @cmfcmf thoughts on this?

Guite:
No idea nor experience with this.

cmfcmf:
same.

If this is really required, it could be improved by querying the time with http://php.net/manual/de/function.microtime.php and sending a header if a second has passed - thus removing the hard sleep.

shefik:
@cmfcmf I think recommending to use microtome is a good idea. I'll modify my code to test the results, and update the pull request accordingly.

craigh:
ping @shefik

craigh:
ping @shefik I need this fixed if you want it included in Core-1.4.2, otherwise it waits for 1.4.3

shefik:
Ok, looking into it.

cmfcmf:
Closing due to lack of response.

PR #2769, opened 2016-02-13
Adding function "modAvailable"

shefik:
Adding function "modAvailable", to check if a module is available in a Twig template.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
+++ b/src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
@@ -80,6 +80,7 @@ class CoreExtension extends \Twig_Extension
             new \Twig_SimpleFunction('setMetaTag', [$this, 'setMetaTag']),
             new \Twig_SimpleFunction('hasPermission', [$this, 'hasPermission']),
             new \Twig_SimpleFunction('adminPanelMenu', [new AdminMenuPanelSimpleFunction($this), 'display'], ['is_safe' => array('html')]),
+            new \Twig_SimpleFunction('modAvailable', [$this, 'modAvailable'])
         );
     }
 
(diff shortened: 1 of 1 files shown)


src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
```diff
@@ -362,4 +363,20 @@ public function hasPermission($component, $instance, $level)
...
     }
+    
+    /**
+     * @param string $modname
+     * @param bool $force
+     * @return bool
+     */
+    public function modAvailable($modname = null, $force = false)
+    {
+        if (empty($modname)) {
```
craigh: if it is not allowed to be empty, then why set it to null above?

craigh:
Mrs must pass all checks before merging. please correct the StyleCI issue.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
shefik: It is set to NULL as a default, to ensure the a module name is always passed. There isn't a specific module name to set as the default.

shefik:
@craigh Fixed.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
craigh: so remove the default value, that will force an argument to be set and you won't have to have code to check that it is set.

shefik:
@craigh The conditional statement and default value has been removed.

craigh:
@shefik slack please.

craigh:
oops - I should have made you add a CHANGELOG for this.

Issue #1753, opened 2014-06-12
Fully implement Twig templating layer


craigh:
Fully implement Twig templating layer

- [x] provide required additional template functions
- [x] theming concept and prototype
- [x] convert system modules
  - [x] Admin #2856
  - [x] Blocks
  - [x] Categories #2923 
  - [x] Extensions
  - [x] Groups #2907 
  - [x] Mailer #2849 
  - [x] PageLock #2862 
  - [x] Permissions #2896 
  - [x] Routes #2921 
  - [x] Search #2853
  - [x] SecurityCenter #2890 
  - [x] Settings #2832 
  - [x] Theme
  - [x] Users #2851
- [x] convert themes
  - [x] Andreas08
  - [x] Atom
  - [x] Bootstrap
  - [x] Printer
  - [x] RSS
  - [x] SeaBreeze
- [x] overall review and bugfixing

Guite:
Closed :smile:

PR #2773, opened 2016-02-16
Refactor ExtensionsModule to Core-2.0 (except Plugin handling).

### LINKED ISSUE #2600: Refactor DisplayHookResponse so that you don't have to use Smarty

craigh:
Refactor ExtensionsModule to Core-2.0 (except Plugin handling).

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | yes
| Tests pass?       | yes
| Fixed tickets     | #2600
| Refs tickets      | #1753
| License           | MIT
| Doc PR            | -
| Changelog updated | yes

 - Upgrade ExtensionsModule to Core-2.0 spec except the handling of module and system plugins.
 - Consolidate all hook-related classes and files into new HookBundle
 - Correct issue with hook display requiring smarty - fixes #2600
 - Add 'info' type of flash message


src/system/ExtensionsModule/Controller/ModuleController.php
```diff
@@ -0,0 +1,193 @@
...
+        $sortableColumns->setOrderByFromRequest($request);
+
+        $vetoEvent = new GenericEvent();
+        $this->get('event_dispatcher')->dispatch(ExtensionEvents::REGENERATE_VETO, $vetoEvent);
+        if (!$vetoEvent->isPropagationStopped() & $pos == 1) {
+            // regenerate the extension list only when viewing the first page
+            $extensionsInFileSystem = $this->get('zikula_extensions_module.bundle_sync_helper')->scanForBundles();
+            $this->get('zikula_extensions_module.bundle_sync_helper')->syncExtensions($extensionsInFileSystem);
+        }
+
```
craigh: ping @Guite above, you can see how I would prefer to handle the coupling with MultiSitesModule. There are other instances in this PR also. Here, MultiSites would simply implement and event handler that does whatever checks it needs and then if it wishes to prevent (veto) the regeneration of extensions (modules) then simply stop propagation of the event.

In current code, this is what the MultiSitesModule is checking for this specific instance (the veto would need to actually invert this logic):

```php
        if ($this->serviceManager['multisites.enabled'] != 1
            || ($this->serviceManager['multisites.mainsiteurl'] == $request->query->get('sitedns', null)
                && $this->serviceManager['multisites.based_on_domains'] == 0)
            || ($this->serviceManager['multisites.mainsiteurl'] == $_SERVER['HTTP_HOST']
                && $this->serviceManager['multisites.based_on_domains'] == 1)) {
```

Guite: So Multisites would also use this `sync helper` for fetching the bundles and process them individually, right?

craigh: no. As I said above, Multisites would use an event handler to veto the regeneration of modules.

src/system/ExtensionsModule/Controller/ModuleController.php
```diff
@@ -0,0 +1,193 @@
...
+            throw new AccessDeniedException();
+        }
+
+        $sortableColumns = new SortableColumns($this->get('router'), 'zikulaextensionsmodule_module_viewmodulelist');
+        $sortableColumns->addColumns([new Column('displayname'), new Column('state')]);
+        $sortableColumns->setOrderByFromRequest($request);
+
+        $vetoEvent = new GenericEvent();
+        $this->get('event_dispatcher')->dispatch(ExtensionEvents::REGENERATE_VETO, $vetoEvent);
+        if (!$vetoEvent->isPropagationStopped() & $pos == 1) {
```
cmfcmf: Typo: `&&`.

src/system/ExtensionsModule/Controller/ModuleController.php
```diff
@@ -0,0 +1,193 @@
...
+                'icon' => 'wrench',
+                'color' => 'black',
+                'title' => $this->__f('Edit %s', ['%s' => $module['displayname']])
+            ];
+        }
+
+        return $actions;
+    }
+
+    private function isCoreModule($moduleName)
```
cmfcmf: I don't like this function.Could this be replaced by `$kernel->getBundle($moduleName) instanceof AbstractCoreModule`?

cmfcmf: Ah maybe this breaks when the module is not installed / inactive (thinking of PageLock here).

craigh: I had something else even than that, but went this this because it is simple.

cmfcmf: Is this the only part of the core where there is a list of system modules? If not, then it's probably better to move it into a static class method somewhere.

PR #2804, opened 2016-03-17
minor cleanup regarding Multisites

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no


src/system/ExtensionsModule/Api/AdminApi.php
```diff
@@ -1465,4 +1452,21 @@ private function getInstaller($module, array $modInfo)
...
+    {
+        $msConfig = $this->serviceManager['multisites'];
+
+        return (
+            ($msConfig['mainsiteurl'] == $this->request->query->get('sitedns', null)
+                && $msConfig['based_on_domains'] == 0)
+            || ($msConfig['mainsiteurl'] == $_SERVER['HTTP_HOST']
+                && $msConfig['based_on_domains'] == 1)
+        );
+    }
```
craigh: This entire `AdminApi` is deprecated and not used by the new Api. IMO it is pointless to change it unless it is preventing functionality (which it should not be). That is why it was left alone. Everything is happening in `ModuleController`, `BundleSyncHelper`, `ExtensionHelper` and `ExtensionStateHelper` now. That is why I have implemented the listeners.

craigh: @Guite you never responded to my concerns here.

craigh:
@Guite I think this should be changed to remove the changes to AdminApi since that function is not used in the core and shouldn't be used outside the core (there would be other problems anyway, I think). If you have found an instance where that methods is used, we should change _that_ to use the new api instead of changing this method.x

Guite:
Changed.

Issue #2805, opened 2016-03-19
Unable to find template "ZikulaThemeModule:Theme:setAsDefault.html.twig".


Kaik:
Error occurs when I tried to change default theme to Bootstrap.
Error is due to template name which is lower case here https://github.com/zikula/core/blob/1.4/src/system/ThemeModule/Resources/views/Theme/setasdefault.html.twig and template annotation here which is camel case https://github.com/zikula/core/blob/1.4/src/system/ThemeModule/Controller/ThemeController.php#L99

I think the easiest way of fixing it would be changing template name to camel case.

Kaik:
Seems like this bug is in extensions as well. I will check where else.

craigh:
Probably my Mac ignoring file casing

PR #2806, opened 2016-03-19
correct casing of template filenames

Kaik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | no
| Fixed tickets     | #2805 
| Refs tickets      | #2805 
| License           | MIT
| Doc PR            | -
| Changelog updated | [no]


Kaik:
Do not merge. it looks like there is more than one occurrence of this bug in other system modules that were recently re-factored to use @Template annotation.

craigh:
status?

craigh:
After looking around at other controllers and templates recently reworked, I think you have found them all.

Issue #2790, opened 2016-03-06
implement CompilerPassInterface 


craigh:
> In Symfony 2.8, this code is no longer needed because if your extensions implement CompilerPassInterface they will be automatically registered and the process() method will be executed during the container compilation.

http://symfony.com/blog/new-in-symfony-2-8-dependencyinjection-improvements#simplified-registering-of-compiler-passes

craigh:
The #2808 PR was merged and then reverted so this ticket is still open.

Guite:
I've added some cleanup in #3150 

Beside this I can not get `LinkContainerPass` (and others) to work. Also I can't find anything about the improvement linked in the original description above in the docs: http://symfony.com/doc/current/components/dependency_injection/compilation.html

Thus I propose to close #2790 as won't fix.

craigh:
http://symfony.com/doc/2.8/components/dependency_injection/compilation.html#execute-code-during-compilation

the blue note below the first code block says
> Prior to Symfony 2.8, extensions implementing CompilerPassInterface were not automatically registered. You needed to register them as explained in the next section.

that is the issue in this ticket.

I am ok with the postponement. I'm not sure I agree with all this auto-wiring, auto-execution anyway. We should look carefully at this before proceeding.

Guite:
Moved to 2.0.0-beta1

craigh:
I don't think we will implement this.

PR #2808, opened 2016-03-21
utilise automatic registration of compiler passes

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | yes
| Fixed tickets     | #2790
| Refs tickets      | -
| License           | MIT
| Doc PR            | -
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/CoreBundle/DependencyInjection/Compiler/DoctrinePass.php
+++ b/src/lib/Zikula/Bundle/CoreBundle/DependencyInjection/Compiler/DoctrinePass.php
@@ -1,4 +1,15 @@
 <?php
+/**
+ * Copyright Zikula Foundation 2015 - Zikula Application Framework
+ *
+ * This work is contributed to the Zikula Foundation under one or more
+ * Contributor Agreements and licensed to You under the following license:
+ *
+ * @license GNU/LGPLv3 (or at your option, any later version).
+ *
+ * Please see the NOTICE file distributed with this source code for further
+ * information regarding copyright and licensing.
+ */
 
 namespace Zikula\Bundle\CoreBundle\DependencyInjection\Compiler;
 
--- a/src/lib/Zikula/Bundle/CoreBundle/CacheClearer.php
+++ b/src/lib/Zikula/Bundle/CoreBundle/CacheClearer.php
@@ -6,7 +6,6 @@
  * Contributor Agreements and licensed to You under the following license:
  *
  * @license GNU/LGPv3 (or at your option any later version).
- * @package Zikula
  *
  * Please see the NOTICE file distributed with this source code for further
  * information regarding copyright and licensing.
--- a/src/lib/Zikula/Bundle/CoreBundle/CoreBundle.php
+++ b/src/lib/Zikula/Bundle/CoreBundle/CoreBundle.php
@@ -1,27 +1,20 @@
 <?php
+/**
+ * Copyright Zikula Foundation 2015 - Zikula Application Framework
+ *
+ * This work is contributed to the Zikula Foundation under one or more
+ * Contributor Agreements and licensed to You under the following license:
+ *
+ * @license GNU/LGPLv3 (or at your option, any later version).
+ *
+ * Please see the NOTICE file distributed with this source code for further
+ * information regarding copyright and licensing.
+ */
 
 namespace Zikula\Bundle\CoreBundle;
 
-use Matthias\SymfonyServiceDefinitionValidator\Compiler\ValidateServiceDefinitionsPass;
-use Symfony\Component\DependencyInjection\Compiler\PassConfig;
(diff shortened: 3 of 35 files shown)


Guite:
This apparently causes problems during the CLI install process (see Travis).: `The service "zikula.doctrine.schema_tool" has a dependency on a non-existent service "doctrine.entitymanager".`
The `doctrine.entitymanager` alias is created by the DoctrinePass class here: https://github.com/zikula/core/blob/1.4/src/lib/Zikula/Bundle/CoreBundle/DependencyInjection/Compiler/DoctrinePass.php#L14
It is not recognised automatically...

Guite:
Probably composer tries to execute stuff before the container has been compiled.

craigh:
why is this advantageous?

Guite:
Sorry I don't understand your comment.

craigh:
why is it better to "utilise automatic registration of compiler passes"? why not the original way?

craigh:
I guess my question seems dumb because I am the one that submitted the original ticket (#2790). I had forgotten I had done so :blush: 

Even so, do you plan to try to get this to work or should we close the PR?

Guite:
I tend to close both the ticket as well as the PR, since it seems we cannot easily fix it (unless somebody has an idea about it)...

craigh:
I think we just need to change the dependency service definition from `doctrine.entitymanager` to the original (non-aliased). can't think of what it is right now. `php app/console debug:container` will give you all the service names. you can look it up.

craigh:
@Guite - rebase this branch/PR please

Guite:
After I changed the service names in several other places the travis build now works.

craigh:
rebase is not the same as merge.

craigh:
locally, please:

  1) rebase to HEAD of `zikula:1.4` branch
  2) soft reset to last core commit
  3) commit all the changes with one commit. Use CONTRIBUTING header as the commit message
  4) force push to `Guite:compilerpassinterface`

If you need assistance with this, I will be happy to walk you through it. I will be available about 20:00 your time.

Guite:
Seems I've coped with it.

craigh:
> Seems I've coped with it.

much better! and thank you. Have you tested locally?

src/lib/Zikula/Bundle/CoreBundle/DependencyInjection/Compiler/DoctrinePass.php
```diff
@@ -1,4 +1,15 @@
 <?php
+/**
+ * Copyright Zikula Foundation 2015 - Zikula Application Framework
```
craigh: all the headers say 2015. It is *2016* now :smile:

craigh: I don't care about this really. eventually we need to replace the header entirely.

cmfcmf:
:+1:

craigh:
@Guite - some quick testing on this branch shows it doesn't work.

screen shot 2016-03-23 at 8 53 00 pm

The rest of the trace is too long to post. I'm not sure what the problem is but it seems to only affect this branch. Please take a look and correct.

Guite:
Can not reproduce locally. Do you remember which specific pages you called?

craigh:
> Do you remember which specific pages you called?

any. I just checked out the branch, manually cleared by cache and every page resulted in that error. :(

Guite:
I did a fresh install using this branch and it worked - at least the pages I called, for example Extensions admin, Mailer admin, admin category showing modules, and some others.

craigh:
I will try again

Guite:
Just the be sure about theory: there is no difference between doing a new install and clearing the cache in an existing one, right? Symfony just recompiles the container which should include the compiler passes anyway.

craigh:
> there is no difference between doing a new install and clearing the cache in an existing one, right?

not entirely correct. There are other artifacts which which can affect the situation. A *closer* approximation to a new install would be to clear all caches (delete the entire directories), clear `/web` (save `.htaccess`), `/vendors`, `/logs`, then removing `app/config/custom_parameters.yml`, `app/config/dynamic/generated.yml`, `config/personal_config.php` and all tables in the DB.

then `composer install`
then install core

craigh:
I completely wiped my installation and started from scratch (did what I said above) and this branch **worked OK** so I guess it is clean to merge. I'm guessing it will require something like that in all dev environments though when merged. Although I don't know why.

craigh:
well, the link collector isn't working anymore in some cases. very confusing.

craigh:
I reset this commit and force pushed to eliminate this merge entirely from the history. It broke the compiler passes, so there is obviously something not working. I shouldn't have merged it so quickly.

PR #2816, opened 2016-03-23
Removed {break} Smarty tag

cmfcmf:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Tests pass?       | ---
| Fixed tickets     | ---
| Refs tickets      | ---
| License           | MIT
| Doc PR            | ---
| Changelog updated | no

[Twig doesn't allow `break`ing a loop](http://twig.sensiolabs.org/doc/tags/for.html#adding-a-condition). In our case, the `break` was there as a slight performance improvement and can safely be removed.


craigh:
thanks. I thought I caught all those.

Issue #2593, opened 2015-09-14
Routing problem caused by speed improvement


Guite:
I got the following error when calling `/mymodule/fooBar/view?lct=admin`:
```
The _controller value "GuiteMyModule:Foobar:View" maps to a "Guite\MyModule\Controller\FoobarController" class, but this class was not found. Create this class or check the spelling of the class and its namespace.
```

The class was actually named `FooBarController` (with a capitalised `B`).

Solved by disabling this line:
https://github.com/zikula/core/pull/2519/files#diff-99827764e017f3db015ca76d386760f1R66

craigh:
@cmfcmf please look into this when you are finishing the Routing PR

craigh:
@cmfcmf can you please look into this when reviewing other Routing stuff

craigh:
@Guite - please test this issue again. that code you reference in your comment has been commented out of the current codebase. (see [here](https://github.com/zikula/core/blob/1.4/src/system/RoutesModule/Routing/RouteLoader.php#L150-L209))

@cmfcmf  - is there some reason you left that code in place and not simply removed it all?

Guite:
This issue can be closed, since default routes are not loaded from database anymore.

craigh:
:+1:

cmfcmf:
> is there some reason you left that code in place and not simply removed it all?

Well, it's still planned to re-enable loading custom routes from the database.

PR #2901, opened 2016-06-22
Allow sub directories in Controller directory

Kaik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | [yes]
| New feature?      | [yes]
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | [no]

## Description
This is a copy paste solution from https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Controller/ControllerNameParser.php#L103

This change allows controllers to be organised in subdirectories

### DIFF
--- a/src/system/RoutesModule/Routing/RouteLoader.php
+++ b/src/system/RoutesModule/Routing/RouteLoader.php
@@ -342,19 +342,14 @@ class RouteLoader extends Loader
      */
     private function sanitizeController($bundleName, $controllerString)
     {
-        if (strpos($controllerString, '::') === false) {
+        if (0 === preg_match('#^(.*?\\\\Controller\\\\(.+)Controller)::(.+)Action$#', $controllerString, $match)) {
             return $controllerString;
         }
-
-        $action = substr($controllerString, strpos($controllerString, '::') + 2);
-        $func = lcfirst(substr($action, 0, -6));
-
-        $a = strrpos($controllerString, '\\') + 1;
-        $b = strrpos($controllerString, '::');
-        $controller = substr($controllerString, $a, $b - $a);
-        $type = substr($controller, 0, -10);
-
-        return $bundleName . ':' . $type . ':' . $func;
+        
+        $controllerName = $match[2];
+        $actionName = $match[3];       
+  
+        return $bundleName . ':' . $controllerName . ':' . $actionName;
     }
 
     /**
--- a/src/system/RoutesModule/Routing/RouteLoader.php
+++ b/src/system/RoutesModule/Routing/RouteLoader.php
@@ -345,11 +345,9 @@ class RouteLoader extends Loader
         if (0 === preg_match('#^(.*?\\\\Controller\\\\(.+)Controller)::(.+)Action$#', $controllerString, $match)) {
             return $controllerString;
         }
-        
-        $controllerName = $match[2];
-        $actionName = $match[3];       
-  
-        return $bundleName . ':' . $controllerName . ':' . $actionName;
+            
+        // Bundle:controller:action
+        return $bundleName . ':' . $match[2] . ':' . $match[3];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


craigh:
ping @cmfcmf

craigh:
it looks like you need to rebase this. Style CI is trying to fix code that @guite fixed already. maybe it doesn't need rebase... dunno.

Kaik:
I will just do another patch and PR - only one question Is there a need to assign matches to vars with names? https://github.com/Kaik/core/blob/37c4ce4314f09b953656f15459b15ee9dc1c82c3/src/system/RoutesModule/Routing/RouteLoader.php#L349

craigh:
> I will just do another patch and PR

please no 🚫 

the code is already fixed. you don't need to fix it.

Kaik:
I don't understand... what I wanted to do is to close this PR and do another same on current 1.4 with @Guite changes already there?

craigh:
> what I wanted to do is to close this PR and do another same on current 1.4

ah. I misunderstood your intention.

I don't think you need to. we can just merge this without the styleCI check. I think it will be ok. You should update your core before further PRs though.

craigh:
I still want @cmfcmf to review this before merging.

Kaik:
It ok now I think...

src/system/RoutesModule/Routing/RouteLoader.php
```diff
@@ -342,19 +342,14 @@ private function prependBundlePrefix(Route $route, AbstractBundle $bundle)
...
-        $b = strrpos($controllerString, '::');
-        $controller = substr($controllerString, $a, $b - $a);
-        $type = substr($controller, 0, -10);
-
-        return $bundleName . ':' . $type . ':' . $func;
+        
+        $controllerName = $match[2];
+        $actionName = $match[3];       
+  
+        return $bundleName . ':' . $controllerName . ':' . $actionName;
```
cmfcmf: The `$actionName` (previously `$func`) now isn't being `lcfirst`-ed anymore. 
I don't know whether this is an issue or not.

Kaik: Well for symfony it should be ok, as this fix is taken from symfony. I do not know how it (lcfirst) will affect legacy. I guess we could add lcfirst here https://github.com/Kaik/core/blob/37c4ce4314f09b953656f15459b15ee9dc1c82c3/src/system/RoutesModule/Routing/RouteLoader.php#L265 so it will be more consistent (instead of ```$func``` lcfirst-ed in one place and ```$type``` in other).
Another thing which can be affected is currently commented out loading routes from DB and #2593 issue.

cmfcmf: >  guess we could add lcfirst here

Sounds good to me.

src/system/RoutesModule/Routing/RouteLoader.php
```diff
@@ -262,7 +262,7 @@ private function setZikulaDefaults(Route $route, AbstractBundle $bundle, $bundle
         $controller = $this->sanitizeController($bundleName, $defaults['_controller']);
         $controller = explode(':', $controller);
         $defaults['_zkType'] = $type = lcfirst($controller[1]);
-        $defaults['_zkFunc'] = $func = $controller[2];
+        $defaults['_zkFunc'] = $func = lcfirst($controller[2]);
```
cmfcmf: :+1:

src/system/RoutesModule/Routing/RouteLoader.php
```diff
@@ -342,19 +342,14 @@ private function prependBundlePrefix(Route $route, AbstractBundle $bundle)
...
-
-        $a = strrpos($controllerString, '\\') + 1;
-        $b = strrpos($controllerString, '::');
-        $controller = substr($controllerString, $a, $b - $a);
-        $type = substr($controller, 0, -10);
-
-        return $bundleName . ':' . $type . ':' . $func;
+        
+        $controllerName = $match[2];
+        $actionName = $match[3];
```
Kaik: How about those assignments, are they needed? I have left them for clarification but returning matches only would be more efficient  ``` return $bundleName . ':' . $match[2] . ':' . $match[3]; ```  I can add comments above for claryfication?

cmfcmf:
Thank you @Kaik!

Issue #2844, opened 2016-04-10
ContextErrorException in Engine.php line 340: Warning: Invalid argument supplied for foreach() in Engine.php line 340


rallek:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.2
| PHP Version       | 5.5.30

### Expected behavior
showing nor error

### Actual behavior
home page is working well but If I go to another page (e.g. site search) I do get the following error:
````
Whoops, looks like something went wrong.
2/2 ContextErrorException in Engine.php line 340: Warning: Invalid argument supplied for foreach()

    in Engine.php line 340
    at ErrorHandler->handleError('2', 'Invalid argument supplied for foreach()', '/home/www/142.koester123.de/Zikula_Core-1.4.2/system/ThemeModule/Engine/Engine.php', '340', array('themeConfig' => null, 'request' => object(Zikula_Request_Http), 'requestAttributes' => array('_theme' => 'ZikulaOmiTheme', '_controller' => 'Zikula\RoutesModule\Controller\RedirectingController::removeTrailingSlashAction', '_zkBundle' => 'ZikulaRoutesModule', '_zkModule' => 'ZikulaRoutesModule', '_zkType' => 'redirecting', '_zkFunc' => 'removeTrailingSlash', 'url' => 'news/view/page/2/', '_route' => 'zikularoutesmodule_redirecting_removetrailingslash', '_locale' => 'de', '_route_params' => array('_zkBundle' => 'ZikulaRoutesModule', 'url' => 'news/view/page/2/', '_locale' => 'de'), '_legacy' => true), 'lct' => null, 'pathInfo' => '/news/view/page/2/')) in Engine.php line 340
    at Engine->setMatchingRealm() in Engine.php line 222
    at Engine->getRealm() in ZikulaVersionDataCollector.php line 37
    at ZikulaVersionDataCollector->collect(object(Zikula_Request_Http), object(Response), object(ContextErrorException)) in Profiler.php line 218
    at Profiler->collect(object(Zikula_Request_Http), object(Response), object(ContextErrorException)) in ProfilerListener.php line 128
    at ProfilerListener->onKernelResponse(object(FilterResponseEvent), 'kernel.response', object(TraceableEventDispatcher))
    at call_user_func(array(object(ProfilerListener), 'onKernelResponse'), object(FilterResponseEvent), 'kernel.response', object(TraceableEventDispatcher)) in WrappedListener.php line 61
[...]

PR #2920, opened 2016-07-12
[WIP] Disable automatic block wrapping for Core 2.0 themes.

Kaik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
This allows for core 2.0 blocks to not be automatically wrapped in div. All html wrapping functions should be deprecated and removed in Core 2.0. To maintain functionality theme developer can use block related variables in block template.

## Todos
- [x] Tests
- [x] Changelog

Not yet finished... @craigh  this is what I think might be a good solution to get rid of those wrappings in core 2.0 completely.

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -4,6 +4,8 @@ CHANGELOG - ZIKULA 1.4.x
 * 1.4.3 (?)
 
  - BC Breaks:
+    - Blocks are no longer wrapped in helper div for Core 2.0 type themes (#2920)
+        - This behaviour was moved to templates. See Andreas08Theme and SeaBreezeTheme for usage examples.
     - Assetic Bundle has been removed (#2939).
     - User block function removed. It is going to be added to the Profile module instead.
     - Old Authentication_Method_Api system has been completely removed.
--- a/src/system/ThemeModule/AbstractTheme.php
+++ b/src/system/ThemeModule/AbstractTheme.php
@@ -54,7 +54,7 @@ abstract class AbstractTheme extends AbstractBundle
 
     /**
      * generate a response wrapped in the theme
-     *   wrap the maincontent in a unique div.
+     *
      * @param string $realm
      * @param Response $response
      * @param string $moduleName
@@ -63,14 +63,14 @@ abstract class AbstractTheme extends AbstractBundle
     public function generateThemedResponse($realm, Response $response, $moduleName = null)
     {
         $template = $this->config[$realm]['page'];
-        $content = '<div id="z-maincontent" class="'
-            . ($realm == 'home' ? 'z-homepage' : '')
-            . (isset($moduleName) ? ' z-module-' . strtolower($moduleName) : '') . '">'
-            . $response->getContent()
-            . '</div>';
-        $response->setContent($content);
+        
+        $templateParameters = [
+            'realm' => (isset($realm) ? $realm : ''),
+            'modulename' => (isset($moduleName) ? $moduleName : ''),
+            'maincontent' => $response->getContent()
+        ];        
 
-        return $this->getContainer()->get('templating')->renderResponse($this->name . ':' . $template, ['maincontent' => $response->getContent()]);
+        return $this->getContainer()->get('templating')->renderResponse($this->name . ':' . $template, $templateParameters);
     }
 
     /**
--- a/src/system/ThemeModule/AbstractTheme.php
+++ b/src/system/ThemeModule/AbstractTheme.php
@@ -81,7 +81,7 @@ abstract class AbstractTheme extends AbstractBundle
      * @param $blockTitle
      * @return string
      */
-    public function generateThemedBlockContent($realm, $positionName, $blockContent, $blockTitle)
+    public function generateThemedBlockContent($realm, $blockType, $bid, $positionName, $blockContent, $blockTitle)
     {
         if (isset($this->config[$realm]['block']['positions'][$positionName])) {
             $template = $this->name . ':' . $this->config[$realm]['block']['positions'][$positionName];
@@ -91,6 +91,9 @@ abstract class AbstractTheme extends AbstractBundle
         }
 
         $templateParameters = [
+            'btype' => strtolower($blockType),
+            'bid' => $bid,
+            'position' => strtolower($positionName),
             'title' => $blockTitle,
             'content' => $blockContent
         ];
--- a/src/system/ThemeModule/AbstractTheme.php
+++ b/src/system/ThemeModule/AbstractTheme.php
@@ -63,12 +63,12 @@ abstract class AbstractTheme extends AbstractBundle
     public function generateThemedResponse($realm, Response $response, $moduleName = null)
     {
         $template = $this->config[$realm]['page'];
-        
+
         $templateParameters = [
             'realm' => (isset($realm) ? $realm : ''),
             'modulename' => (isset($moduleName) ? $moduleName : ''),
             'maincontent' => $response->getContent()
-        ];        
+            ];
 
         return $this->getContainer()->get('templating')->renderResponse($this->name . ':' . $template, $templateParameters);
     }
--- a/src/system/ThemeModule/Engine/Engine.php
+++ b/src/system/ThemeModule/Engine/Engine.php
@@ -145,7 +145,7 @@ class Engine
         if (!isset($activeTheme) || !$activeTheme->isTwigBased()) {
             return false;
         }
-        
+
         $bid = !empty($blockInfo['bid']) ? $blockInfo['bid'] : 'none';
         $blockType = !empty($blockInfo['blocktype']) ? $blockInfo['blocktype'] : 'none';
         $position = !empty($blockInfo['position']) ? $blockInfo['position'] : 'none';
--- a/src/system/ThemeModule/Engine/Engine.php
+++ b/src/system/ThemeModule/Engine/Engine.php
@@ -145,8 +145,11 @@ class Engine
         if (!isset($activeTheme) || !$activeTheme->isTwigBased()) {
             return false;
         }
+        
+        $bid = !empty($blockInfo['bid']) ? $blockInfo['bid'] : 'none';
+        $blockType = !empty($blockInfo['blocktype']) ? $blockInfo['blocktype'] : 'none';
         $position = !empty($blockInfo['position']) ? $blockInfo['position'] : 'none';
-        $content = $activeTheme->generateThemedBlockContent($this->getRealm(), $position, $blockInfo['content'], $blockInfo['title']);
+        $content = $activeTheme->generateThemedBlockContent($this->getRealm(), $blockType, $bid, $position, $blockInfo['content'], $blockInfo['title']);
 
         return $content;
     }
@@ -166,7 +169,7 @@ class Engine
     {
         if (!$legacy) {
             // legacy blocks are already themed at this point. @todo at Core-2.0 remove $legacy param and this check.
-            $content = $this->getTheme()->generateThemedBlockContent($this->getRealm(), $positionName, $content, $title);
+            return $this->getTheme()->generateThemedBlockContent($this->getRealm(), $blockType, $bid, $positionName, $content, $title);
         }
 
         // always wrap the block (in the previous versions this was configurable, but no longer) @todo remove comment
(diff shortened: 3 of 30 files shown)


craigh:
things you must test:
 - [ ] Twig-based block in Smarty-based theme
 - [ ] Twig-based block in Twig-based theme
 - [ ] Smarty-based block in Smarty-based theme
 - [ ] Smarty-based block in Twig Based theme
 - [ ] bootstrap based themes
 - [ ] non-bootstrap based themes

Kaik:
I will do those tests, no problem. I posted it quite quickly but I will go thru it all and mark functions that should be deprecated for Core 2.0 as well as write docs. Additionally, I will look for a way to move this https://github.com/Kaik/core/blob/b983cf378b25f4a501c62b42f246a867a5d516d6/src/system/ThemeModule/AbstractTheme.php#L66 somewhere so BC will be maintained, but for Core 2.0 themes we should allow theme developer to decide if he want's this div or not - basically by passing related variables to corresponded template.
This way I believe we could get rid of html from those php files completely in Core 2.0

craigh:
@paustian @Kaik I'm still not clear why wrapping content (of any kind) in a `<div>` is a problem. Isn't the problem that this div has *styling* (e.g. css), not that there is an actual div. Taking a block-level element and wrapping it in another block-level element should have no effect. right? So - maybe the solution is to change the css?

Kaik:
Well, there are two kinds of problems with it. First yes that div element may sometimes break some layouts I do not have any example with me now, but I remember having, a problem with it a long time ago (Postune 0.7xx) - I think it was positioning or something like it... can't remember right now.

Second is that we have HTML in our PHP in 2016 when we have templating systems this is a very bad idea to keep it as Core 2.0 standard. Even allowing to disable it somewhere as it was done before is a bad idea unnecessary complication this thing should be done in a template.

We can pass variables that are used to make that div to the related template and let theme developer decide if he wants to have this kind of div in his layout or not, he can as well do exactly same thing there or something similar but with different class names, do conditions basing on those variables etc.. much more freedom than having fixed div in your template you can't change, turn off etc.. long time ago it was fancy feature and still there was an option to turn it off now we cannot even do that.

Kaik:
@craigh I just checked 1.3.x themes and they do not work... what I mean by that is:
1. Something in UsersModule LogInBlock I will create, a ticket.
2. While block wrapping work for 1.3.x themes content wrapping does not work at all, and this is not an issue with changes I will make a ticket.

Kaik:
This is done and can be merged. There are errors left with legacy which are not caused by this change.
1. Something in UsersModule LogInBlock. #2934 
2. While block wrapping work for 1.3.x themes content wrapping does not work at all, and this is not an issue with changes in this PR. #2935

While I have not checked 1. I have traced 2 and this PR works as expected. (I have temporary fixed #2935 by hardcoding toplevelmodule to true)

craigh:
I will review and potentially merge this after I return from traveling for work in about 10 days.

Kaik:
:+1:

craigh:
With respect, I am still suspicious of this solution. If we are forced to adjust core themes (as you just did) then isn't it, by definition, *not* BC? Every theme that is out there becomes "broken"? Maybe I do not understand...

As I said above in my very first comment, every type of theme must be compatible (and tested) in order to maintain BC. I don't get the impression that that will be true if this PR is merged. To be fair, I have not yet pulled this locally to evaluate it more closely.

In theory, I agree with your concerns about _forward compatibility_, in that we do not want to mix layout and content, etc. I'm just not sure this is the right approach.

Kaik:
Hi, @craigh I'm just reminding about that my writing English is bad and I might have hard to express what I want to say. 
>With respect, I am still suspicious of this solution. If we are forced to adjust core themes (as you just did) then isn't it, by definition, not BC? 

Hmm for me BC term is mostly regarding 1.3.x themes and modules. This PR deals only with new 1.4.1+ twig themes and it is not affecting 1.3.x themes at all. BC breaks in this PR regards only new twig themes.
>Every theme that is out there becomes "broken"? Maybe I do not understand...

It is possible that few new themes since twig themes were introduced are basing on those divs, especially ones that are basing on Andreas or SeaBreeze but I do not think there are many of them. I think most new twig based themes are based on Bootstrap theme and there is a possibility that some one used those div classes. 

>As I said above in my very first comment, every type of theme must be compatible (and tested) in order to maintain BC. I don't get the impression that that will be true if this PR is merged. To be fair, I have not yet pulled this locally to evaluate it more closely.

As I mentioned above this affects only new Twig themes.

>In theory, I agree with your concerns about forward compatibility, in that we do not want to mix layout and content, etc. I'm just not sure this is the right approach.

There is another variation of this solution. We could, in theory, create template and place it in theme module and call it from engine so we can override it etc.

This comment is not finished...

craigh:
> There is another variation of this solution. We could, in theory, create template and place it in theme module and call it from engine so we can override it etc.

I had also considered this solution. An advantage here is that a core template can be provided (which includes the wrapping divs) and then the theme could override that template if desired.

I am looking more closely at the PR at the moment, but I keep running into other core bugs which prevent an accurate review...

Another thought that I am thinking over is the original purpose of the wrapping divs and their need going forward... As far as I can tell, the divs are there to provide a standard naming convention and tools (via css) to affect layout consistently in all modules, etc. This is especially true in the admin side (for example: [this](https://github.com/zikula/core/blob/83add14510bb84af21472edfc1508d83d30fcab9/src/style/legacy.css#L54))

The example above is in the legacy.css file, which obviously implies that it will be gone. I cannot event remember at the moment when it is loaded and when it is not, etc. 

The overriding question then is "should the core provide layout for certain aspects of the system?"  @cmfcmf @Guite And I am not sure of the answer. If the core does not do so at all, then the theme designer has to do more, but has more flexibility, I suppose.

craigh:
> I keep running into other core bugs which prevent an accurate review...

I fixed some ([1](https://github.com/zikula/core/commit/24ade5200b52080d0054deaf7e0580feac533852) & [2](https://github.com/zikula/core/commit/46bf01ca3c01e5e018fc56a6b973689cd791f2f4)) so now I can see some of this.

here is a _minor_ but important issue:

1.4 branch:
screen shot 2016-08-09 at 3 52 31 pm

this PR branch:
screen shot 2016-08-09 at 3 53 04 pm

craigh:
> there is already huge BC break between 1.4.0 and 1.4.1 themes

maybe so. please refresh my memory. I cannot remember what they are?

Kaik:
>maybe so. please refresh my memory. I cannot remember what they are?

This won't fix anyway but it is an example that we cannot have full BC to new thigs we introduce  #2844 because we are working on it.

>here is a minor but important issue:

This should be fixed now I thought Bootstrap module do not use legacy .z-maincontent classes but it does in admin.

craigh:
screen shot 2016-08-10 at 2 53 29 pm
screen shot 2016-08-10 at 2 53 39 pm

craigh:
for your information:
[LegacySeaBreezeTheme.zip](https://github.com/zikula/core/files/411886/LegacySeaBreezeTheme.zip)

craigh:
It occurs to me that I should have compared those images to similar images from the 1.4 branch. I will do that in a few hours.

craigh:
Top is legacy SeaBreeze on the 1.4 branch. bottom is 'normal' SeaBreeze on the 1.4 branch.
screen shot 2016-08-10 at 5 05 21 pm
screen shot 2016-08-10 at 5 04 59 pm

craigh:
so - it appears that they are mostly identical. (which is good). There are some minor differences in the non-legacy versions: 
 - the block titles are styled a bit differently (see "Search box" and "User log-in" in the second picture on each.
 - the spacing after "Cal" (the block title) is different (again, second picture on each example)

I'm not sure these differences are worth discussion but I point them out as small changes.

Kaik:
Ok @craigh more important is to check if there are related classes in source because it might be a css issue and it might be not related to this PR. I will join you I need to refresh my branches now because something went wrong when I rebase yesterday...

Kaik:
Hmm something is very wrong I just have rebased then reinstalled and now I'm checking everything and for some reason, one block is rendered twice in helper div... Andreas08Theme 
screenshot from 2016-08-11 00-28-39

Other block is rendered properly... hmm I need to go sleep now I will investigate it more tomorrow.

craigh:
also true in a quick check here using bootstrap theme.
screen shot 2016-08-10 at 6 55 32 pm

craigh:
but not true in standard 1.4 branch.
screen shot 2016-08-10 at 6 57 10 pm

craigh:
here (this is still bootstrap theme - which is twig based) you can see that the behavior is different for the calendar block (smarty) than for the account links block (twig)
screen shot 2016-08-10 at 7 02 15 pm

however, the account links block is still wrong (so they are both wrong in different ways). The account links block wrapper should _include_ the title (e.g. `<h3>account links</h3>`) but it does not.

craigh:
In a smarty based theme (in this case the LegacySeaBreezeTheme) the html is correct for both block types using your branch.
screen shot 2016-08-10 at 7 11 04 pm

Kaik:
There are two issues one is wrapping. I have made content wrapped not the whole block my mistake.
here and probably in few other blocks https://github.com/Kaik/core/blob/f4a401145ca11e9400fc061a4f984ec3a3657eb0/src/themes/BootstrapTheme/Resources/views/Block/block.html.twig#L4

Second is related to legacy block's but I need to trace it somewhere there is logic problem and I do not know where yet.

Kaik:
I have traced it a bit but I do not know if this issue wasn't present before... need to do more test...
This is how it goes.
1. ``` showblockposition('topnav') ``` end up showing block here https://github.com/zikula/core/blob/1.4/src/system/BlocksModule/Twig/Extension/BlocksExtension.php#L149
There are only two types of block's those which use BlockHandlerInterface (legacy false) and \Zikula_Controller_AbstractBlock ( legacy true) Second one is already wrapped in unique div!

2. $this->themeEngine->wrapBlockContentInTheme is called for both block's https://github.com/zikula/core/blob/1.4/src/system/ThemeModule/Engine/Engine.php#L165
this is the place where things are getting messed up when block is legacy and is already wrapped it us wrapped again here 
```
      if (!$legacy) {
            // legacy blocks are already themed at this point. @todo at Core-2.0 remove $legacy param and this check.
            $content = $this->getTheme()->generateThemedBlockContent($this->getRealm(), $positionName, $content, $title);
        }
        // always wrap the block (in the previous versions this was configurable, but no longer) @todo remove comment
        return $this->wrapBlockContentWithUniqueDiv($content, $positionName, $blockType, $bid);
```

So it is a bug

Kaik:
Ok I have found the problem... this PR affects it... This was expected actually hmm 
...Thinking aloud...
We need to make legacy work as it is working while we want to change the way new things work.

If we assume that Smarty themes are legacy and Twig themes are new ones we could provide a solution which is not ideal beacuse it requires people that already moved to Twig themes to add unique div in their block templates if they want to use them.
 
**Solutin one** 
If theme is Smarty based we provide unique div wrapper inside php as it is done now, legacy to 1.3.x is maintained and we can remove it in Core 2.0. 

If theme is Twig based we are not wrapping any kind of block ( BlockHandlerInterface or \Zikula_Controller_AbstractBlock  ) we just pass unwrapped content to block template and do wrapping there.

Pros:
We are on right track with new things we provide 1.3.x feature with higest flexibility we can have unique div can be used or not on, modified and this will be adjustable on block position basis - for example all left blocks are wrapped all right not or are wrapped in different unique div.
Cons:
Already converted Twig themes need to be adjusted (unique div has to be added in case someone want to use it)

**Solution two**
Unigue div core template we mentioned this before. Separate template is called in engine and is doing the wrapping user can override it in his theme.

Pros:
Already converted Twig themes not need to be adjusted
Cons: 
Not that flexible as solution one template can be overriden but that is working for all blocks. I personally do not like it. It adds additional complication and reduces flexibility.

Both solutions keep 1.3.x BC it is only the way new things work in Twig theme.

@craigh  @cmfcmf  @Guite ?

Guite:
Is this related to #2935?

Kaik:
No

Guite:
Personally I'd prefer a separate (theme-independent) template which decorates the main content with that div. This would avoid that every theme needs to add this div redundantly.

Issue #2685, opened 2015-12-17
Drop Assetic


Guite:
Since we use zasset (which is built above the Symfony Asset component) we should remove Assetic from the core.

craigh:
since it is currently included since 1.4.0 we should perhaps schedule removal for 2.0.0?

Guite:
Wouldn't this lead to more people starting using it? I think at the moment nobody is using Assetic (yet), so it could be advantageous getting rid of it before it becomes adopted?

craigh:
yes, that would be a concern, but we have much tech in the Core that is @deprecated. We leave it for BC. Of course there is no way to know that "nobody is using it" (until we remove and someone complains :wink: ).

Guite:
Postponed issue to 2.0.0.

craigh:
[here](https://github.com/symfony/symfony-standard/pull/860) is a conversation about removing ascetic from symfony-standard (which is not the same as 'symfony', apparently). just FTR

PR #2939, opened 2016-07-19
Remove AsseticBundle

Kaik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | [no]
| New feature?      | [yes/no]
| BC breaks?        | no well...
| Deprecations?     | no
| Fixed tickets     |  #2685
| Refs tickets      |  #2685
| License           | MIT
| Changelog updated | [yes]

## Description
I have read #2685 and I think we should remove it before anyone will start using it before it's too late.


Guite:
:+1:

craigh:
I have the same reservations still. what if someone is using it?

Kaik:
I really doubt that there is anyone who is using it and know what is doing. Even if there is a person or two who are using it on purpose and they know what it does, well they should be able to install it again. Anyone else should use zasset.

craigh:
please add changelog in the BC Break area.

Issue #2840, opened 2016-04-03
[1.4.2] Installation issues to fix


craigh:
from http://zikula.org/forums/topic/59974

| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.2
| PHP Version       | 5.x unknown

### Expected behavior

full install with no errors

### Actual behavior
 - [x] Needed to install Apache module access_compat to process order commands in htaccess files. Without this no javascript or styles loaded into install page.
 - [x] Link on installation to documentation on install page didn't work. Page looked for INSTALL-1.4.0.MD. Changed the name of the file in docs/en to match and then link worked.
 - [x] Installation progressed quickly through all steps except the last (Install assets) which reported that it failed. Then could not do login. Error was "Data too long for column 'sessid' at row 1" in table session_info. I changed the length of this field from 40 to 60. I think the session id was 52 characters long. Login then succeeded and installation appeared complete.

craigh:
PHP Version 5.5.14
Apache 2.4.16
OpenSuse distribution Leap 42.1 in Virtual Box

craigh:
refs #2938

cmfcmf:
> Needed to install Apache module access_compat to process order commands in htaccess files. Without this no javascript or styles loaded into install page.

Here is an example of a Symfony `.htaccess` file: https://github.com/symfony/symfony-standard/blob/master/src/.htaccess
```htaccess
<IfModule mod_authz_core.c>
    Require all denied
</IfModule>
<IfModule !mod_authz_core.c>
    Order deny,allow
    Deny from all
</IfModule>
```

craigh:
therefore my comment above:
>refs #2938

you are saying that changing the .htaccess file like that would also solve the `access_compat` issue?

cmfcmf:
> you are saying that changing the .htaccess file like that would also solve the access_compat issue?

No, just wanted to add a reference to Symfony. I haven't actually looked in depth and I don't know whether this fixes the actual problem.

Guite:
First point (#2938) fixed in #2982

Guite:
Second point fixed in #2983

craigh:
thanks @Guite nice work.

PR #2972, opened 2016-08-16
increase length of session_id column in session table. refs #2840

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no maybe?
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | #2840 
| License           | MIT
| Changelog updated | yes

## Description
lengthen the column to allow for longer session ids. addresses concern raised on forum and outlined again in #2840 

I'm concerned that this will truncate the session table, forcing all to login again. My local tests are inconclusive. @cmfcmf @Guite @Kaik thoughts or testing?

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -16,6 +16,7 @@ CHANGELOG - ZIKULA 1.4.x
  - Fixes:
     - Fix error on creation of new ExtendedMenublock.
     - Fix display of blocks using theme overrides (#2872).
+    - Lengthen SesssionId column in session table (#2840).
 
  - Features:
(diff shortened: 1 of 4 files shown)


cmfcmf:
> I'm concerned that this will truncate the session table, forcing all to login again. My local tests are inconclusive.

Do you consider having to login again a big issue?

Guite:
Despite that I can not see how the table should be truncated just by altering (even extending) a column.

craigh:
> Do you consider having to login again a big issue?

not necessarily, but it is an inconvenience, I guess.

> Despite that I can not see how the table should be truncated just by altering (even extending) a column.

Doctrine does some weird things and especially when using the SchemaTool to do _anything_.

Issue #2938, opened 2016-07-19
How to change the htaccess for Apache 2.4


trichers:
In Zikula are some htaccess files to protect the directory:

```
deny from all
<FilesMatch "\.(css|js|png|gif|jpg|bmp)$">
order allow,deny
allow from all
</filesmatch>
```

From Apache 2.4 the 

> Order deny,allow
> Deny from all

must be changed to

> Require all denied

How would be the ne notation for the code above?

Kaik:
Apache 2.2
```
deny from all
<FilesMatch "\.(css|js|png|gif|jpg|bmp)$">
order allow,deny
allow from all
</filesmatch>
```
Apache 2.4
```
Require all denied
<FilesMatch "\.(css|js|png|gif|jpg|bmp)$">
Require all granted
</filesmatch>
```

craigh:
Is this a core issue? is there something that needs to be changed in the code? Or is this simply a support request and now solved?

Guite:
This is a core issue. We need to change all .htaccess files to make them work with both Apache 2.2 as well as Apache 2.4.

Example from https://wordpress.org/support/topic/htaccess-fix-for-apache-22-and-24

```
	# Apache 2.2
	<IfModule !mod_authz_core.c>
		Deny from all
	</IfModule>

	# Apache 2.4
	<IfModule mod_authz_core.c>
		Require all denied
	</IfModule>
```

craigh:
from the topic you linked, it appears Wordpress implemented `<FilesMatch...` instead of the suggested `<Files...`. What do you recommend?

Guite:
Fixed in #2982

PR #2982, opened 2016-08-19
enhanced .htaccess files for supporting both Apache 2.2 and 2.4

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #2938 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/app/cache/.htaccess
+++ b/src/app/cache/.htaccess
@@ -1,2 +1,14 @@
-Order deny,allow
-deny from all
+# ------------------------------------------------------------------
+# Purpose of file: block any web access to unallowed files
+#                  stored under the /app/cache directory
+# ------------------------------------------------------------------
+
+# Apache 2.2
+<IfModule !mod_authz_core.c>
+    Deny from all
+</IfModule>
+
+# Apache 2.4
+<IfModule mod_authz_core.c>
+    Require all denied
+</IfModule>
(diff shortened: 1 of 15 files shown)


src/config/.htaccess
```diff
@@ -1,5 +1,27 @@
-deny from all
+# ------------------------------------------------------------------
+# Purpose of file: block any web access to unallowed files
+#                  stored under the /config directory
+# ------------------------------------------------------------------
+
+# Apache 2.2
+<IfModule !mod_authz_core.c>
+    Deny from all
```
cmfcmf: Shouldn't this also include `Order deny,allow`?

cmfcmf: I see, you removed `Order deny,allow` everywhere so this is a consistent change. I assume `Order deny,allow` is the default.

Guite: I basically didn't change the 2.2 rules at all, but only wrapped them into the `IfModule` check.

src/app/cache/.htaccess
```diff
@@ -1,2 +1,14 @@
-Order deny,allow
```
cmfcmf: You removed this one, that's why I noticed it.

Guite: Good point! Will readd it immediately.

Issue #2884, opened 2016-05-22
Unable to delete categories - Integrity constraint violation


shefik:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.2
| PHP Version       | 5.6

### Expected behavior

Using the UI, delete a category here:

`/categories/admin/view`

### Actual behavior

Error thrown:

```
An exception occurred while executing 'DELETE FROM categories_category WHERE id = ?' with params [42]: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`database_name`.`categories_category`, CONSTRAINT `FK_D0B2B0F8727ACA70` FOREIGN KEY (`parent_id`) REFERENCES `categories_category` (`id`))
```

### Steps to reproduce

Install the Pages module, and then try to delete the parent category "Pages", along with its subcategories.

Guite:
Any idea how to approach this @zikula/contributors

cmfcmf:
@shefik You wrote Zikula version 1.**3**.2. I assume you mean 1.**4**.2?

shefik:
@cmfcmf Yes, I did mean 1.4.2.

craigh:
> Any idea how to approach this?

this is an issue where the foreign key entities have to be removed before the 'parent'. Or a cascade must be set to do so automatically. (And then called properly from doctrine).

I had noticed this behavior before and forgot to do anything about it.

Guite:
Proposal: when a category should be deleted do the following:
1. iterate over all registries
2. for each registry check if the corresponding subtree contains this category
3. if yes check if the mapping table contains a reference to this category
4. if yes abort with an error

cmfcmf:
> Proposal: when a category should be deleted do the following:
> 1. iterate over all registries

I've never used categories so I don't fully grasp how they work, but your proposal sounds alright.

Guite:
Modules register category registries pointing to category subtrees. This allows having different sets of categories, but also reusing them across different modules.

Guite:
This needs also applied to _move_ actions.

Guite:
Fixed in #2985

shefik:
This issue does not appear to be fixed in Zikula 1.4.3. I am experiencing the same error:

[...]

PR #2985, opened 2016-08-19
prevent deleting and moving categories which are used

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #2884 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/lib/util/CategoryUtil.php
+++ b/src/lib/util/CategoryUtil.php
@@ -53,10 +53,10 @@ class CategoryUtil
         $checkCat = self::getCategoryByPath("$rootPath/$name");
         if (!$checkCat) {
             $cat = new \Zikula\CategoriesModule\Entity\CategoryEntity();
-            $em = ServiceUtil::get('doctrine.entitymanager');
-            $em->persist($cat);
+            $entityManager = ServiceUtil::get('doctrine.entitymanager');
+            $entityManager->persist($cat);
             $data = [];
-            $data['parent'] = $em->getReference('ZikulaCategoriesModule:CategoryEntity', $rootCat['id']);
+            $data['parent'] = $entityManager->getReference('ZikulaCategoriesModule:CategoryEntity', $rootCat['id']);
             $data['name'] = $name;
             $data['display_name'] = [$lang => $displayname];
             $data['display_desc'] = [$lang => $description];
@@ -67,7 +67,7 @@ class CategoryUtil
             $data['path'] = "$rootPath/$name";
 
             $cat->merge($data);
-            $em->flush();
+            $entityManager->flush();
             $cat['ipath'] = "$rootCat[ipath]/$cat[id]";
             if ($attributes && is_array($attributes)) {
                 foreach ($attributes as $key => $value) {
@@ -75,7 +75,7 @@ class CategoryUtil
                 }
             }
 
-            $em->flush();
+            $entityManager->flush();
 
             return $cat->getId();
         }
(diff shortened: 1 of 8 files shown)


craigh:
good morning. 🌞  I would like to review this before merging 😄

src/lib/util/CategoryUtil.php
```diff
@@ -492,12 +492,13 @@ public static function deleteCategoryByID($cid)
...
 
-        $dql = "DELETE FROM Zikula\CategoriesModule\Entity\CategoryEntity c WHERE c.id = :cid";
+        // delete attributes first to avoid error about foreign key constraints
+        $dql = "DELETE FROM Zikula\CategoriesModule\Entity\CategoryAttributeEntity a WHERE a.category = :cid";
         $query = $em->createQuery($dql);
         $query->setParameter('cid', $cid);
         $query->getResult();
 
-        $dql = "DELETE FROM Zikula\CategoriesModule\Entity\CategoryAttributeEntity a WHERE a.category = :cid";
+        $dql = "DELETE FROM Zikula\CategoriesModule\Entity\CategoryEntity c WHERE c.id = :cid";
```
cmfcmf: I think the two queries should be in a single transaction. This would fix any ordering issues and will make sure the database remains consistent, even if the second step (deleting the CategoryEntities) fails.

src/system/CategoriesModule/CategoriesModuleInstaller.php
```diff
@@ -76,6 +76,7 @@ public function install()
         $this->setVar('allowusercatedit', 0);
         $this->setVar('autocreateusercat', 0);
         $this->setVar('autocreateuserdefaultcat', 0);
+        $this->setVar('permissionsall', 0);
```
craigh: adding a new modvar means you probably need an upgrade routine as well.

src/system/CategoriesModule/GenericUtil.php
```diff
@@ -169,4 +170,72 @@ public static function processCategoryAttributes($category, $attrib_names, $attr
...
+                    ->findBy(['category' => $category['id']]);
+                if (count($mappings) > 0) {
+                    // existing reference found
+                    return false;
+                }
+            }
+        }
+
+        return true;
+    }
```
craigh: I would prefer that we do not create more static classes (`Util`). But rather non-static services using DIC (in other modules usually called `Helper`). Then the repository can be injected instead of using yet another @deprecated `\ServiceUtil::` call. We either do it now or redo it later....

src/lib/util/CategoryUtil.php
craigh: If entity is properly set up, manual DQL would not be required but simply cascade when using `->remove()`

src/system/CategoriesModule/CategoriesModuleInstaller.php
Guite: It isn't a new modvar, it just was missing in the installer.

src/system/CategoriesModule/GenericUtil.php
Guite: Well, this `GenericUtil` class existed before, just added another method to it. Of course it needs to be refactored to using dependency injection instead, but I'll leave this for another iteration for this module.

src/lib/util/CategoryUtil.php
Guite: I wouldn't change this now, but when `CategoryUtil` is changed as a whole.

craigh: I like @cmfcmf suggestion to wrap them both in one transaction

src/system/CategoriesModule/CategoriesModuleInstaller.php
craigh: ✅

src/system/CategoriesModule/GenericUtil.php
craigh: agreed 👍

PR #2986, opened 2016-08-19
Fix minor issues with the legacy mailer.

cmfcmf:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | ---
| Refs tickets      | ---
| License           | MIT
| Changelog updated | no

The function must return a boolean and not throw exceptions.


craigh:
not sure if I would call this API "legacy" or not. but otherwise 👍

Issue #2914, opened 2016-07-08
require.config.js is not built properly.


Kaik:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.x
| PHP Version       | 5.6

### Expected behavior
Proper baseUrl to components in require.config.js

### Actual behavior
baseUrl is set to "components" (default) 

### Steps to reproduce
Install Zikula and try to use require.js with require.config.js
```
    <script src="{{ asset('require.js') }}"></script>
    <script src="{{ asset('require.config.js') }}"></script> 
    <script>
        require(['jquery'
        ], function ($) {
        }
        );
    </script>  
```

I have tried to use default require.js config created by composer-installers which is present in /web folder but baseUrl to components is wrong this is mostly because it is not configured in our composer.json or should be configured in different place more information's here https://github.com/RobLoach/component-installer#base-url
Changing baseUrl simple to web does not solve this problem as it uses "current url" as part before /web so "http://www.domain.com/en/whatever/web" the only way it works is to use ``` Zikula.Config.baseURL + "web" ```

The second attempt was to use require-buld.js but something was wrong when using installed Backbone

I have ended up using custom require config inside my module main.js file
```
// Require.js allows us to configure shortcut alias
require.config({
    // The shim config allows us to configure dependencies for
    // scripts that do not call define() to register a module
    shim: {
        underscore: {
            exports: '_'
        },
        backbone: {
            deps: [
                'underscore',
                'jquery'
            ],
            exports: 'Backbone'
        }
    },
    paths: {
        jquery: Zikula.Config.baseURL + 'web/jquery/jquery',
        underscore: Zikula.Config.baseURL + 'web/underscore/underscore',
        backbone: Zikula.Config.baseURL + 'web/backbone/backbone'
    }
});
```
I will try to investigate it further. I'm leaving this issue here for now maybe you have idea how can we solve this.

craigh:
> Install Zikula and try to use require.js with require.config.js

are these libraries currently part of the core?

Kaik:
This is installed by roblach component installer and configured automatically.

Kaik:
Best way would be to use ``` Zikula.Config.baseURL + "web" ``` as baseUrl but I do not know if this is possible to do in composer.json. The other way might be to override somehow, somewhere https://github.com/RobLoach/component-installer/blob/master/src/ComponentInstaller/Process/RequireJsProcess.php

Guite:
[...]

PR #2995, opened 2016-08-19
Add custom regeneration functions for requirejs.

Kaik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | [yes]
| New feature?      | [no]
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #2914 
| Refs tickets      | #2914 
| License           | MIT
| Changelog updated | [no]

## Description
This enables custom baseUrl.

### DIFF
--- a/src/lib/Zikula/Composer/Process/RequireJsProcess.php
+++ b/src/lib/Zikula/Composer/Process/RequireJsProcess.php
@@ -229,7 +229,7 @@ EOT;
         $merged = $array1;
 
         foreach ($array2 as $key => &$value) {
-            if (is_numeric($key)){
+            if (is_numeric($key)) {
                 $merged[] = $value;
             } else {
                 if (is_array($value) && isset($merged[$key]) && is_array($merged[$key])) {
--- a/src/lib/Zikula/Composer/Process/RequireJsProcess.php
+++ b/src/lib/Zikula/Composer/Process/RequireJsProcess.php
@@ -49,9 +49,8 @@ class RequireJsProcess extends Process
         // Construct the require.js and stick it in the destination.
         $json = $this->requireJson($this->packages, $this->config);
         $requireConfig = $this->requireJs($json);
-        $thisPath = dirname(__DIR__);
-        $vendorPath = str_replace("lib/Zikula/Composer", "vendor/robloach/component-installer/src/ComponentInstaller", $thisPath);
-        //var_dump($vendorPath);
+        $vendorPath = str_replace("lib/Zikula/Composer", "vendor/robloach/component-installer/src/ComponentInstaller", dirname(__DIR__));
+
         // Attempt to write the require.config.js file.
         $destination = $this->componentDir . '/require.config.js';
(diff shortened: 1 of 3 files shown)


Kaik:
This is raw not cleaned. It does work but I do not know if you like this solution.

craigh:
if you click the [Details](https://styleci.io/analyses/qBAxNg) link next to StyleCI check, you can then login at the top and then you should be able to submit a PR to yourself to automatically fix all the StyleCI issues.

src/lib/Zikula/Composer/Process/RequireJsProcess.php
```diff
@@ -0,0 +1,268 @@
...
+     * {@inheritdoc}
+     */
+    public function process()
+    {
+        // Construct the require.js and stick it in the destination.
+        $json = $this->requireJson($this->packages, $this->config);
+        $requireConfig = $this->requireJs($json);
+        $thisPath = dirname(__DIR__);
+        $vendorPath = str_replace("lib/Zikula/Composer", "vendor/robloach/component-installer/src/ComponentInstaller", $thisPath);
+        //var_dump($vendorPath);
```
craigh: remove

src/lib/Zikula/Composer/Process/RequireJsProcess.php
```diff
@@ -0,0 +1,268 @@
...
+     */
+    public function requireJs(array $json = array())
+    {
+        // Encode the array to a JSON array.
+        $js = JsonFile::encode($json);
+
+        // Construct the JavaScript output.
+        $output = <<<EOT
+var components = $js;
+components.baseUrl = Zikula.Config.baseURL + "web";
```
craigh: what if Zikula.js is not included?

craigh:
> I do not know if you like this solution

I think we are all happy with ANY solution at this point. 😄

src/lib/Zikula/Composer/Process/RequireJsProcess.php
Kaik: It will fail of course. https://github.com/zikula/core/blob/7b1c8fdeaade4cf5e85c0843425351314712e163/src/javascript/zikula/zikula.boot.js#L9

craigh:
Is Zikula.js "strict" JavaScript or is it Scriptaculous?

craigh:
Is this finished?

Kaik:
@craigh We barely use javascript in Zikula there are maybe two places where we have more complicated scripts that change color or hide something. Some basic system information's are needed for more complex javascript at the moment we have Zikula object that provides that information when this will change I do not know but I'm kind of confident that base URL will be included and then we might need to change this.

craigh:
> Is Zikula.js "strict" JavaScript or is it Scriptaculous? or other? Prototype...

- [ ] yes
- [ ] no

> Is this finished?

- [ ] yes
- [ ] no

Kaik:
Yes it is finished. Zikula.js is strict but under this namespace, some parts may use Scriptaculous

Issue #3031, opened 2016-08-23
An exception has been thrown during the rendering of a template ("Notice: Undefined index: url") in @ZikulaAdminModule/AdminInterface/header.html.twig at line 9.


robbrandt:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.3 build 157
| PHP Version       | 5.5.26

### Expected behavior
PostCalendar upgrade should run without error

### Actual behavior
[1] Twig_Error_Runtime: An exception has been thrown during the rendering of a template ("Notice: Undefined index: url") in "@ZikulaAdminModule/AdminInterface/header.html.twig" at line 9.
    at n/a
        in /Users/robbrandt/Documents/htdocs/zcs2/vendor/twig/twig/lib/Twig/Template.php line 402

    at Twig_Template->displayWithErrorHandling(array('caller' => array('_theme' => 'ZikulaAndreas08Theme', 'pos' => '1', '_controller' => 'Zikula\ExtensionsModule\Controller\ModuleController::viewModuleListAction', '_zkBundle' => 'ZikulaExtensionsModule', '_zkModule' => 'ZikulaExtensionsModule', '_zkType' => 'module', '_zkFunc' => 'viewModuleList', '_locale' => 'en', '_route' => 'zikulaextensionsmodule_module_viewmodulelist', '_route_params' => array('pos' => '1', '_zkBundle' => 'ZikulaExtensionsModule', '_locale' => 'en'), '_template' => object(Template)), 'domain' => 'zikula', 'app' => object(AppVariable), 'pagevars' => object(ParameterBag), 'themevars' => object(ParameterBag), 'currentUser' => object(CurrentUserApi)), array())
        in /Users/robbrandt/Documents/htdocs/zcs2/vendor/twig/twig/lib/Twig/Template.php line 355

[...]

PR #3034, opened 2016-08-25
Fix admin footer / header

cmfcmf:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | ---
| Refs tickets      | ---
| License           | MIT
| Changelog updated | no

Call the Controllers instead of simply including the templates, as they might (and do!) define different template variables.

### DIFF
--- a/src/system/AdminModule/Resources/views/Admin/footer.html.twig
+++ b/src/system/AdminModule/Resources/views/Admin/footer.html.twig
@@ -1,1 +1,1 @@
-{{ include('@ZikulaAdminModule/AdminInterface/footer.html.twig') }}
+{{ render(controller('ZikulaAdminModule:AdminInterface:footer')) }}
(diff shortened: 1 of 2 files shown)


craigh:
seems ok? I guess I don;t really understand. this was @Kaik 's thing.

Kaik:
This is ok you could as well use ```{{ adminHeader() }} ``` Twig tag.
I think this will not solve #3031

craigh:
ah yes. For consistency, let's use `{{ adminHeader() }}` and `{{ adminFooter() }}` ?

craigh:
@Guite this is from your refactoring. Did you do similar methods in other places or did you use the twig functions?

Kaik:
This will be changed anyway to fix #3031. I need to rewrite this to use AdminController::categorymenuAction. I will redirect all smarty tags to use controller refactored by @Guite these support legacy. Then add ```deprecated``` to all functions and files that need to be removed when smarty is removed.

craigh:
@Kaik are you saying close this PR and you will fix in a new one?

Kaik:
yes please

cmfcmf:
The same issues still exists.

PR #3036, opened 2016-08-25
Improvements to 7.0 compat

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
updates for improved compatibility with PHP 7 and minor improvements to avoid Session usage when using CLI.

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxUpgradeController.php
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxUpgradeController.php
@@ -55,6 +55,7 @@ class AjaxUpgradeController extends AbstractController
     public function commandLineAction($stage)
     {
         $this->container->setParameter('upgrading', true);
+
         return $this->executeStage($stage);
     }
 
@@ -62,8 +63,6 @@ class AjaxUpgradeController extends AbstractController
     {
         switch ($stageName) {
             case "loginadmin":
-                $this->yamlManager->setParameter('upgrading', true);
-
                 return $this->container->get('core_installer.controller.ajaxinstall')->loginAdmin();
             case "upgrademodules":
                 $result = $this->upgradeModules();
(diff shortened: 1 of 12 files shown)


src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxUpgradeController.php
```diff
@@ -188,9 +188,11 @@ private function from141to142()
             return true;
         }
         // do some clean up
-        \SessionUtil::delVar('interactive_init');
-        \SessionUtil::delVar('interactive_remove');
-        \SessionUtil::delVar('interactive_upgrade');
+        if ($this->container->get('request_stack')->getCurrentRequest() && $this->container->get('request_stack')->getCurrentRequest()->hasSession()) {
```
cmfcmf: Uuhm. Maybe introduce a variable for the current request? :boat:

src/system/RoutesModule/Twig/TwigExtension.php
```diff
@@ -135,4 +135,32 @@ public function displayPathAsString($path, $route)
...
+    {
+    }
+
+    public function getImageThumb($params)
+    {
+    }
+
+    public function profileLink($uid, $class = '', $maxLength = 0)
+    {
+    }
```
cmfcmf: Why are these needed :question:

craigh: they call session related functions and get instantiated on install causing the php7 build to fail

craigh: and they are generated functions that are not needed in the core.

cmfcmf: Oh I see. They overwrite the functions in the base class.

craigh: yes 😀

Guite: `getUserAvatar` is used here:
https://github.com/zikula/core/blob/1.4/src/system/RoutesModule/Resources/views/Helper/includeStandardFieldsDisplay.html.twig#L18
https://github.com/zikula/core/blob/1.4/src/system/RoutesModule/Resources/views/Helper/includeStandardFieldsDisplay.html.twig#L25
So you might remove that as well.

PR #3037, opened 2016-08-25
override symfony native RoutingExtension for cases where route is not…

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | [yes/no]
| New feature?      | [yes/no]
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT

## Description
A few sentences describing the overall goals of the pull request's commits.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog

… available. @cmfcmf

### DIFF
--- a/src/system/ExtensionsModule/Resources/config/services.xml
+++ b/src/system/ExtensionsModule/Resources/config/services.xml
@@ -15,6 +15,7 @@
         <parameter key="zikula_extensions_module.extension_dependency_repository.class">Zikula\ExtensionsModule\Entity\Repository\ExtensionDependencyRepository</parameter>
         <parameter key="twig.extension.zikula_extensions_module.class">Zikula\ExtensionsModule\Twig\Extension\ExtensionsExtension</parameter>
         <parameter key="twig.extension.zikula_extensions_module.user_interface.class">Zikula\ExtensionsModule\Twig\Extension\UserInterfaceExtension</parameter>
+        <parameter key="twig.extension.routing.class">Zikula\ExtensionsModule\Twig\Extension\RoutingExtension</parameter>
         <parameter key="zikula_extensions_module.bundle_sync_helper.class">Zikula\ExtensionsModule\Helper\BundleSyncHelper</parameter>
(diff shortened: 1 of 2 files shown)


craigh:
@cmfcmf here is an idea

cmfcmf:
I disagree with this. I think the default behaviour of these functions should remain.

craigh:
> I disagree with this. I think the default behaviour of these functions should remain.

solution then is to use 

`{% if modAvailable('MyCustomModule') %}{{ path('mycustommodule_controller_method') }}{% endif %}`

for all routes that could possibly not be available at time of upgrade, etc.

ping @shefik

craigh:
see https://github.com/zikula/core/commit/18233ac856620dce54112bff92a29c127daf4cdd

shefik:
@craigh I did that already, wrapping the route in the conditional statement. But, it still resulted in the same error.

craigh:
interesting.

craigh:
I've tested this and it works as I expected.

How I tested it:

I added this code to the top of `src/system/ExtensionsModule/Resources/views/Module/viewModuleList.html.twig`

    Profile: {{ modAvailable('ZikulaProfileModule') }} : {{ isBundle('ZikulaProfileModule') }}
    {% if modAvailable('ZikulaProfileModule') %}{{ path('zikulaprofilemodule_user_index') }}{% endif %}

the `isBundle()` function is a temporary plugin/function I added just to check if the kernel has the bundle loaded. It only does this `return $this->container->get('kernel')->isBundle($name);` and the results were always the same as `modAvailable()`

When profile was uninstalled entirely, it looked like
`Profile: :`

when installed and active it looked like
`Profile: 1 : 1 /core.git/src/profile/`

when installed and inactive it looked like
`Profile: :`

when installed but requiring upgrade it looked like
screen shot 2016-08-27 at 8 56 13 am
`Profile: 1 : 1 /core.git/src/profile/`

and the upgrade completed and worked as expected.

@shefik wrote:
> I did that already, wrapping the route in the conditional statement. But, it still resulted in the same error.

I cannot explain the behavior of your module(s). Every test I have run works as I expect.

PR #3065, opened 2016-09-02
Improve Forward Compatibility

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
changed several areas of code to ensure FC. In general, we must all avoid use of any old static class, e.g. `ModUtil`, `System`, `ZLanguage` etc. In some cases, this is currently impossible (`ZLanguage` and `PageUtil` are problems). But solutions must be found and FC code must be used wherever possible.


src/lib/bootstrap.php
```diff
@@ -36,4 +36,7 @@
 $kernel->boot();
 
 // legacy handling
-require __DIR__.'/core.php';
+//require __DIR__.'/core.php';
+//ServiceUtil::setContainer($kernel->getContainer());
+//EventUtil::setManager($kernel->getContainer()->get('event_dispatcher'));
+//ModUtil::initCoreVars();
```
craigh: these extra lines are just part of the testing I was doing. Many things work with only the set up of those three items and not the full boot up of the old core

craigh:
@Guite @cmfcmf @Kaik and anyone with an opinion. I would like to merge this now. speak up if you have objections.

Kaik:
I would like to say that a cheat sheet would be helpful. A single file where static classes to service name are listed. I can help creating it.

Guite:
Another way to document it would be `@deprecated` tags.

craigh:
> I can help creating it.
great!

PR #3077, opened 2016-09-12
provide shortcut method for retrieving system variables

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description
This adds a new method to the variable api which allows easier retrieval of system variables. In particular importing of the `VariableApi` becomes unrequired, because the `CONFIG` constant is not used in the calling class anymore.


Guite:
Opinions about that?

@craigh @cmfcmf @Kaik

Kaik:
I have mixed feelings for shortcuts like this one why not use ``->get('ZConfig', ....)`` ? I'm not really against or something. Maybe add ``hasSystemVar()`` and ``getAllSystemVars()`` as well for consistency?

craigh:
I'm not sure I agree. I didn't do this in the first place because I didn't want it to be 'too easy' to access/modify system vars.

Additionally, if this is truly desired, before merging you would need to
 - add `hasSystemVar` `getAllSystemVars` `setSystemVar` `setAllSystemVars` `delSystemVar` `delAllSystemVars` methods
 - add tests for all methods.
 - add Twig functions for all

Then comes the task of refactoring everything to use this.

Guite:
> why not use ->get('ZConfig', ....)

To avoid that a hardcoded string value is used everywhere.

> Maybe add hasSystemVar() and getAllSystemVars() as well for consistency?

Sure, this could make sense. But `hasSystemVar()` should always return `true`, shouldn't it? And `getAllSystemVars()` is probably never required anywhere.

> add `hasSystemVar getAllSystemVars setSystemVar setAllSystemVars delSystemVar delAllSystemVars` methods

I wouldn't allow setting and deleting system vars from arbitrary modules. What modules typically do is reading and using them, but they do not change them (except for rare special cases).

Kaik:
>To avoid that a hardcoded string value is used everywhere.

Well, I do not think that this value will change anytime soon, but I understand your concern. If we want to be consistent with avoiding that string we would have to add what @craigh said so there will be no string usage in any possible case even deleting all system vars method. 

> I wouldn't allow setting and deleting system vars from arbitrary modules. What modules typically do is reading and using them, but they do not change them (except for rare special cases).

This is true but when rare special case will occur it will break avoiding string value so either we use it for all cases or we do not use it at all.

Guite:
> when rare special case will occur it will break avoiding string value

You can use the `CONFIG` constant in these cases.

cmfcmf:
> This adds a new method to the variable api which allows easier retrieval of system variables. In particular importing of the VariableApi becomes unrequired, because the CONFIG constant is not used in the calling class anymore.

I am 👍 with this change.

craigh:
please add a twig function then. I guess I can live with it. I won't vote against it.

Guite:
Added Twig function and tests.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
```diff
@@ -365,6 +366,20 @@ public function getModVar($module, $name, $default = null)
...
 
     /**
+     * @param $name
+     * @param null $default
+     * @return mixed
+     */
+    public function getSystemVar($name, $default = null)
+    {
+        if (empty($name)) {
+            throw new \InvalidArgumentException(__('Empty argument at') . ':' . __FILE__ . '::' . __LINE__);
```
cmfcmf: Why does the exception contain the file name and line? This is redundant information.

cmfcmf:
I am :+1: except for the exception message.

src/lib/Zikula/Bundle/CoreBundle/Twig/Extension/CoreExtension.php
cmfcmf: Also, it shouldn't use the legacy __() translation.

Guite: It was copied from `getModVar` so there it is wrong, too. Going to look at it.

Guite: Why is file name and line redundant?

Guite:
> task of refactoring everything to use this

done for the core (added to this PR)

PR #3086, opened 2016-09-25
Correcting pagevars with language translation.

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
Correcting pagevars with language translation.

### DIFF
--- a/src/system/ThemeModule/EventListener/DefaultPageVarSetterListener.php
+++ b/src/system/ThemeModule/EventListener/DefaultPageVarSetterListener.php
@@ -61,8 +61,8 @@ class DefaultPageVarSetterListener implements EventSubscriberInterface
         $this->pageVars->set('langdirection', 'ltr'); //\ZLanguage::getDirection());
         $this->pageVars->set('title', $this->variableApi->getSystemVar('defaultpagetitle_'.\ZLanguage::getLanguageCode()));
         $this->pageVars->set('meta.charset', $this->getCharSet());
-        $this->pageVars->set('meta.description', $this->variableApi->getSystemVar('defaultmetadescription'));
-        $this->pageVars->set('meta.keywords', $this->variableApi->getSystemVar('metakeywords'));
+        $this->pageVars->set('meta.description', $this->variableApi->getSystemVar('defaultmetadescription_'.\ZLanguage::getLanguageCode()));
+        $this->pageVars->set('meta.keywords', $this->variableApi->getSystemVar('metakeywords_'.\ZLanguage::getLanguageCode()));
         $this->pageVars->set('homepath', $this->router->generate('home'));
         $this->pageVars->set('coredata', ['version' => \Zikula_Core::VERSION_NUM]); // @todo
     }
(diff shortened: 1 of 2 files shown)


craigh:
you must correct [issues that do not pass StyleCI](https://styleci.io/analyses/XZQKaR)

src/system/ThemeModule/EventListener/DefaultPageVarSetterListener.php
```diff
@@ -59,10 +59,10 @@ public function setDefaultPageVars(GetResponseEvent $event)
         // set some defaults
         $this->pageVars->set('lang', $event->getRequest()->getLocale()); //\ZLanguage::getLanguageCode());
         $this->pageVars->set('langdirection', 'ltr'); //\ZLanguage::getDirection());
-        $this->pageVars->set('title', $this->variableApi->getSystemVar('defaultpagetitle'));
+        $this->pageVars->set('title', $this->variableApi->getSystemVar('defaultpagetitle_'.\ZLanguage::getLanguageCode()));
```
craigh: I am not in favor of adding more usage of `ZLanguage` to Core-2.0 (like this listener). you can see that I have already replaced some of it. You can use `$event->getRequest()->getLocale()` instead

src/system/ThemeModule/EventListener/DefaultPageVarSetterListener.php
```diff
@@ -59,10 +59,10 @@ public function setDefaultPageVars(GetResponseEvent $event)
...
         $this->pageVars->set('meta.charset', $this->getCharSet());
-        $this->pageVars->set('meta.description', $this->variableApi->getSystemVar('defaultmetadescription'));
-        $this->pageVars->set('meta.keywords', $this->variableApi->getSystemVar('metakeywords'));
+        $this->pageVars->set('meta.description', $this->variableApi->getSystemVar('defaultmetadescription_'.\ZLanguage::getLanguageCode()));
+        $this->pageVars->set('meta.keywords', $this->variableApi->getSystemVar('metakeywords_'.\ZLanguage::getLanguageCode()));
```
craigh: same as above.

craigh:
Also - these changes are **supposed to be** unneeded. Somewhere (can't look now) when the CoreVariables are loaded, it is supposed to translate them already.

craigh:
> Somewhere (can't look now) when the CoreVariables are loaded, it is supposed to translate them already. 

This is the method I was referring to:
https://github.com/zikula/core/blob/f8028568435b63126bc9c20df2a110dde67a1fa1/src/lib/util/ModUtil.php#L150-L164

you can see that it is called a few times when the core is booted up (in legacy mode). Also, this should already be taken care of for Core-2.0 methods here: https://github.com/zikula/core/blob/7c6eccb5beb0183ad7ad2e2e61c0bbcc0e90e6a3/src/system/ExtensionsModule/Api/VariableApi.php#L104-L111

So - the correct way to fix this (if it is truly not working) would be to troubleshoot those areas, not add more code on top of it, I think. Will you please take a look there.

Issue #3121, opened 2016-10-23
remove Mapstraction library


craigh:
Mapstraction doesn't seem to have a general use-case but is rather specific. If required, it should be provided by the module that requires it. In addition, it is quite old and left undeveloped for some time now.

https://github.com/mapstraction/mxn

I would like to remove it entirely at Core-2.0 and not replace the functionality at all (as will be done with the other core plugin - Imagine)

opinions?

robbrandt:
I think a common map library would be a good thing though. Leaflet might be a good alternative.

Guite:
While modules can express their own dependencies using composer, in my opinion the core should provide solutions for recurring requirements to avoid redundant libraries wasting disk space. I see three points:

1. Images
Imagine definitely needs to be central. Many many modules need it.

2. Maps
Interesting for every module working with some kind of addresses (for example calendars) or other geo location functionality.
Yes, leaflet is a very good replacement for mapstraction and also recommended by @gfr (who has good experience in geo applications). See also https://github.com/Guite/MostGenerator/issues/340 which is exactly about replacing mapstraction by leaflet in the generator, which is blocked by that leaflet is not in the core.

3. PDF
Some tool for creating PDF documents. For example https://github.com/SlikNL/DompdfBundle for http://dompdf.github.io/ or https://github.com/KnpLabs/KnpSnappyBundle -> see http://jayroman.com/blog/how-to-export-any-symfony-route-to-pdf also

craigh:
I agree with Images and PDF. I don't see Maps as a general use case.

gfr:
I would really suggest to add the leaflet library to the core as maps are often part of modern websites (e.g. Where to find us etc.)

Portugao:
I'm with @gfr .

Guite:
Just an additional idea for bundle integration:
As many Symfony bundles require configuration on application level (`/app/config`) is would possibly be helpful providing a dedicated module (for example named `BundleIntegrationModule`) which users can use to add arbitrary bundles. This module would be almost empty, it's purpose would just be holding bundles and their configuration in order to provide their functionality to the system.

craigh:
I don't think this will work since adding bundles requires manual manipulation of the `ZikulaKernel` and doing this will require the config be in the zikula config. I'm not sure there is a better way at the moment.

craigh:
mapstraction was removed from the master branch some time ago. therefore closing this. I still feel any replacement should not be in the core but rather in the module that requires it.

[...]

Issue #2945, opened 2016-07-22
Evaluate LiipImagineBundle for Imagine system plugin


Guite:
https://github.com/liip/LiipImagineBundle
http://symfony.com/doc/current/bundles/LiipImagineBundle/index.html

craigh:
I think this should be done.

I also think the concept of a "System Plugin" should just be replaced by bundles that can be added to the core's composer/kernel.

Issue #1948, opened 2014-08-23
implement Namespaced Plugin architecture


craigh:
The Plugin architecture is still entirely using Pear classes and doesn't appear to have any Namespaced alternatives.

cmfcmf:
> but I do not understand why it is called a Module Plugin instead of simply a Plugin

There are system plugins and module plugins. Scribite is an example using module plugins.

Guite:
@craigh at #2945:

> I also think the concept of a "System Plugin" should just be replaced by bundles that can be added to the core's composer/kernel.

Guite:
Just saw that I said the same above :smile:

craigh:
I think Module plugins (like scribite editors, etc) should be replaced by [Symfony Tagged Services](http://symfony.com/doc/current/service_container/tags.html). The module that utilizes the plugins would be responsible to manage them and provide a UI for config, etc.

SO - in order for this ticket to close, we simply need to document that all plugins are deprecated - make sure all related classes are marked as such and move them (if not already) into the legacy folder.

cmfcmf:
:+1:

Issue #1644, opened 2014-03-05
Add thumb cleanup interval configuration to Imagine plugin


espaan:
The Imagine plugin has the option to automatically cleanup the generated thumbnails every day. This datetime should be configurable. This can be for instance twice per day or only once every week for less visited sites.

Guite:
This is a won't fix, because the imagine plugin is deprecated and being replaced by LiipImagineBundle. http://symfony.com/doc/current/bundles/LiipImagineBundle/index.html

PR #3122, opened 2016-10-23
Imagine

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | yes - ImaginePlugin
| Fixed tickets     | #2945, #1948
| Refs tickets      | #1644, #3121, #703, #577, #418
| License           | MIT
| Changelog updated | yes

## Description
Add, install and configure the LiipImagineBundle. **Deprecate all plugins**.

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -7,6 +7,9 @@ CHANGELOG - ZIKULA 1.4.x
 
  - Deprecated:
    - ExtMenu and MenuTree blocks are deprecated and will not be refactored to Twig/Core-2.0.
+   - All System plugins and Module plugins and the systems supporting them are deprecated.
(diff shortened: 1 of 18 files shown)


cmfcmf:
How would a module add more filter sets?

craigh:
> How would a module add more filter sets?

it would not. I don't know how to make that happen. ideas?

craigh:
> How would a module add more filter sets?

of course, it can always use _runtime options_ (http://symfony.com/doc/current/bundles/LiipImagineBundle/basic-usage.html#runtime-options)

cmfcmf:
> of course, it can always use runtime options

Right, I forgot those. This could work then 😄

PR #3138, opened 2016-11-05
[FEATURE] implement asset merger management. closes #2912

### LINKED ISSUE #2912: "Enable CSS/JS combination" Not Working
| Q | A |
| --- | --- |
| Zikula Version | 1.4.x |
| PHP Version | 5.6 |
### Expected behavior

CSS and JS is expected to be combined, with GZ and CSS minification enabled.
### Actual behavior

CSS and JS are not combined. Using a Twig-based theme.
### Steps to reproduce

Try to enable the settings at `theme/admin/config`.

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #2912 
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description
[FEATURE] implement asset merger management. closes #2912

### DIFF
--- a/src/app/config/parameters.yml
+++ b/src/app/config/parameters.yml
@@ -33,7 +33,11 @@ parameters:
 
     zikula.session.name: _zsid
     assets_version: ~ #e.g. 'v1'
-    
+    zikula_asset_manager.combine: true
+    zikula_asset_manager.lifetime: "1 day"
+    zikula_asset_manager.compress: false
+    zikula_asset_manager.minify: false
+
     # Changing the current umask to the octal "022" will set directory permissions
     # to "755", and file permissions to "644" (for directories such as
     # "app/cache" and "app/logs"). Changing the current umask to the octal "02"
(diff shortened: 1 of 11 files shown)


src/app/config/parameters.yml
```diff
@@ -33,7 +33,11 @@ parameters:
 
     zikula.session.name: _zsid
     assets_version: ~ #e.g. 'v1'
-    
+    zikula_asset_manager.combine: true
+    zikula_asset_manager.lifetime: "1 day"
+    zikula_asset_manager.compress: false
+    zikula_asset_manager.minify: false
+
```
Guite: Why not reusing the existing module vars (which would be more consistent for the user)?

craigh: because I don't think it should be *that* dynamic. This isn't a feature that needs to be enabled/disabled a bunch of times by the user. Additionally, using parameters separates the legacy methods from the new.

Issue #3052, opened 2016-08-29
Set up mail service on install


craigh:
| Q | A |
| --- | --- |
| Zikula Version | 1.4.3-rc2 |
| PHP Version | 5.6.24-1+deb.sury.org~xenial+1 |

Perhaps the installer should attempt to set up the mail service or at least notify the admin that the mail service requires set up.

craigh:
additionally, if the email server is not setup it should be handled better than a 500 - this was after attempting to register with the email service not properly set up

![screen_shot_2016-08-28_at_11 58 46_am](https://cloud.githubusercontent.com/assets/350048/18049032/b79990c4-6db3-11e6-8a34-48cf0b1189ca.png)

MrMontesa:
I just stumbeld upon this issue since I'm not able to get mailing feature to work.
In case mailing cant be setup during installation process I would recommend to set it to "DEV - no mailing" and highlight this with an info window in admin area.

Guite:
The problem about the 500 error has already been fixed in 5872d6c545dc6ebbc2ccef9b4502350d74ea89e2

Guite:
Any proposal about the actual feature request: how should the installer do that? Adding an additional step offering mail configuration options to setup after the user entered his admin credentials?

MrMontesa:
From end-user perspective I would propose the follwing:
- no Zikula function, setting change, etc should end in a 500 and break the system.
- On install, set the mail feature to "DEV - no mail"
- In admin area display an info windows as long as mailing feature is disabled. Same like "There is a new Zikula version available"
- The info windows should contain a direct link into the mail config area to change the settings

craigh:
> Perhaps the installer should attempt to set up the mail service or at least notify the admin that the mail service requires set up.

I think the installer should *try* to set up the mailer on its own first. Maybe it can try several methods and see if they are available somehow.

I do not think the mailer should be disabled or set to DEV on install. This will mean the product is not functional "out of the box" and this will only lead to support requests.

If it determined (somehow) that the mail service cannot be set up/enabled, then a notice should be set in the admin panel with a link as @MrMontesa suggested.

nmpetkov:
I hope this issue is right place to write the problem I have. I attempt to send test email from mailer module and get:
``` It looks like the message could not be sent properly. ```
and the mail is not sent.

Please point me how to change  configuration to be able to send test mail.

- It is Zikula 1.4.4.build100 fresh install
[...]

PR #3149, opened 2016-11-11
use dev mailer by default; show alert if mailer is not enabled

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3052 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/app/config/dynamic/default.yml
+++ b/src/app/config/dynamic/default.yml
@@ -7,7 +7,7 @@ jms_i18n_routing:
     locales: [en]
     strategy: prefix_except_default
 swiftmailer:
-    transport: mail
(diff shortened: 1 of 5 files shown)


craigh:
do not merge this until I have some time

craigh:
so this disables all mail on install? I'm not sure I agree. 😕

Guite:
This does what was suggested in #3052 after I asked about how it should be done.

craigh:
I disagree with this solution. 👎

PR #3150, opened 2016-11-11
use original service name instead of alias

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | #2790 
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxInstallController.php
+++ b/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxInstallController.php
@@ -241,7 +241,7 @@ class AjaxInstallController extends AbstractController
      */
     private function updateAdmin()
     {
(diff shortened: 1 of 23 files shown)


craigh:
why do you feel this renaming is important?

craigh:
> why do you feel this renaming is important?

I see it is relevant to #2790.

I think some real-usage testing is in order with this PR since the last time caused problems (although I can't see why this should cause issues). So please actually test an install and general usage with this PR and report.

Guite:
I tested this already. This PR does not change any behaviour, it just uses the original service name to reduce dependencies to our custom `DoctrinePass` class.

craigh:
> I tested this already.

:+1: merge it then.

PR #3152, opened 2016-11-11
removed overwrite of RouterListener, fixed #2657

### LINKED ISSUE #2657: We should not overwrite the RouterListener
When I ran the composer update to Symfony 2.8, I received the following:

> Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::installAssets
>   [Matthias\SymfonyServiceDefinitionValidator\Exception\InvalidServiceDefinitionsException]  
>   Service definition validation errors (1):                                                                                                                          
> - router_listener: Argument for type-hint "Symfony\Component\Routing\RequestContext" points to a service of class "Symfony\Component\HttpFoundation\RequestStack"  
> 
> Script Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::installAssets handling the post-update-cmd event terminated with an exception
>   [RuntimeException]  
>   An error occurred when executing the "'assets:install '\''src/web'\'''" command:                                                           
> 
>  [Matthias\SymfonyServiceDefinitionValidator\Exception\InvalidServiceDefinitionsException]                                                
> 
>  Service definition validation errors (1):                                                                                                
> - router_listener: Argument for type-hint "Symfony\Component\Routing\RequestContext" points to a service of class "Symfony\Component\HttpFoundation\RequestStack"

You have a note in the listener about a 3.0 change, but maybe the change has been migrated to 2.8. Can you please look into it?

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #2657 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
First I compared our `RouterListener` class with the original class of Symfony 2.8 because I wanted to see what we are doing differently in order to think about another way of injecting this additional behaviour.
It turned out that we are basically only adding this:
```
unset($parameters['_zkModule'], $parameters['_zkType'], $parameters['_zkFunc']);
```
I thought that this could also be done by an additional listener which is called *after* the RouterListener. But then I just disabled our `RouterListener` and things still worked fine. I also tested a legacy module with this (Content) which also seems still working.


craigh:
@cmfcmf this will require your approval please.

cmfcmf:
I am not exactly sure why we even have the RouterListener. It was introduced by @drak three years ago.
I am :+1: if you tested this :)

(https://github.com/zikula/core/commit/4a77237f42fc6770ab799a3eb74c83c67251cee2)

PR #3153, opened 2016-11-11
typehint to Symfony Request class, fixed #2115

### LINKED ISSUE #2115: Check how we override the Request
We still have the old Zikula_Request thing around, make sure we override the request the proper way: http://symfony.com/doc/current/components/http_foundation/introduction.html#overriding-the-request

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #2115 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
There are a few other occurences, but these are all within `/lib/legacy` anyway.


cmfcmf:
This does not fix the referenced issues.

Guite:
The referenced issue isn't a real issue imo. The legacy subclasses work as they are and are to be removed.

craigh:
@cmfcmf is correct that it doesn't fix the referenced issue. Whether the issue is valid or not is a different question.

Guite:
Sure. No reason to not merge this PR though. Just close #2115 as won't fix then.

craigh:
something like this was already done in master. I'm not looking forward to merging 😢 but I suppose it can be done. I would like @cmfcmf 's approval before merging.

PR #3155, opened 2016-11-11
centralise bs modal fix for mmenu, fixed #3098

### LINKED ISSUE #3098: Find better solution for conflict between mmenu and bs modals
#2903 and #3097 are currently fixed by a workaround which is implemented in the corresponding theme. Better would be a central solution which avoids this problem for other themes.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3098 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no


craigh:
It seems you've fixed legacy stuff here but not Core-2.0 stuff. You must fix both before merging.

Guite:
I don't understand what you are meaning with `Core-2.0 stuff`. The new JS file is included in both `categories.panel.html.twig` and `modules.panel.html.twig`. Also I tested the modals in the permissions module to verify it is loaded.

craigh:
> I don't understand what you are meaning

I meant `system/AdminModule/Resources/views/AdminInterface/categories.panel.html.twig` which you have indeed modified. I didn't see that.

I had just seen changes to the legacy smarty plugin and not seen parallel changes in the Twig function, but apparently you have done so.

So, if all that's true, then I guess this is good to go.

Issue #3177, opened 2016-11-14
No authentication mapping record for admin


Guite:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.4-git
| PHP Version       | 7

### Expected behavior
Admin can use lost password functionality like other users.

### Actual behavior
The user is not found (neither by username nor by email address).

### Steps to reproduce
Try to start a lost password process.
The `zauth_authentication_mapping` table does not contain an entry for the admin user after a fresh installation. After I inserted a record manually into the database the process starts properly.

craigh:
the record is migrated when the admin logs in the first time after the new system.

craigh:
so this is not specific to the admin. This issue will occur to anyone that hasn't logged in since the upgrade past 1.4.3 and forgotten their password. Your new system will need a BC layer for this possibility.

Guite:
Also for new installations?

craigh:
yes.

we could back port code from master if you wish I suppose.

craigh:
https://github.com/zikula/core/blob/1.4/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxInstallController.php#L247

craigh:
https://github.com/zikula/core/blob/master/src/lib/Zikula/Bundle/CoreInstallerBundle/Controller/AjaxInstallController.php#L245-L262

Guite:
looks reasonable for me. any objections about backporting this?

craigh:
no objection

PR #3178, opened 2016-11-14
easier password reset process, fixed #1781, plus some minor fixes

### LINKED ISSUE #1781: Password Reset without Confirmation Code
Most sites just send a "password" reset email to the email address on file, and the user clicks the link to create a password and that's it - the user can then select a new password.

We currently have a two-step password reset with the additional confirmation code. Is a confirmation code necessary? Or, can we just go with a password reset email sans the confirmation code?

Perhaps it can be an option for the site admin to choose whether they want password reset to require a confirmation code, or not.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes (email related)
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #1781 
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

**Notes:**

This does still use the verification entity and confirmation codes. So for example the expiry is still handled by that.

But there are some essential behavioural differences:

1. The code is wrapped into an encoded identifier together with the user id, the user name and the email address. This identifier is appended to the url contained in the email text.
2. After extracting the identifier arguments the user is selected using the user id.
3. The code is checked before the form even displays.
4. The form contains only two password fields now.
5. A new helper class (`LostPasswordVerificationHelper`) encapsulates the details, making the controllers a bit more slim.

This PR also addresses some minor bugs I've found, like non-absolute urls in the email templates and wrong usage of the `MailerApi` in the `MailHelper` class.

### DIFF
--- a/src/system/ZAuthModule/Controller/AccountController.php
+++ b/src/system/ZAuthModule/Controller/AccountController.php
@@ -186,24 +186,13 @@ class AccountController extends AbstractController
             return $this->redirectToRoute($redirectToRoute);
         }
 
-        $user = $this->get('zikula_users_module.user_repository')->findBy([
-            'uid' => $requestDetails['userId'],
-            'uname' => $requestDetails['userName'],
-            'email' => $requestDetails['emailAddress']
-        ]);
-
-        if (count($user) > 1) {
-            $this->addFlash('error', $this->__('There are too many users registered with that address. Please contact the system administrator for assistance.'));
-
-            return $this->redirectToRoute($redirectToRoute);
-        } elseif (count($user) < 1) {
+        /** @var UserEntity $user */
+        $user = $this->get('zikula_users_module.user_repository')->find($requestDetails['userId']);
+        if (null === $user) {
             $this->addFlash('error', $this->__('User not found. Please contact the system administrator for assistance.'));
 
             return $this->redirectToRoute($redirectToRoute);
         }
-
-        /** @var UserEntity $user */
-        $user = $user[0];
 
         if (!$lostPasswordVerificationHelper->checkConfirmationCode($user->getUid(), $requestDetails['confirmationCode'])) {
             $this->addFlash('error', $this->__('Your request could not be processed due to invalid arguments. Maybe your link is expired?'));
--- a/src/system/ZAuthModule/Controller/AccountController.php
+++ b/src/system/ZAuthModule/Controller/AccountController.php
@@ -13,6 +13,7 @@ namespace Zikula\ZAuthModule\Controller;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\RouterInterface;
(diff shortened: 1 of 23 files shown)


Guite:
ready to review

src/system/ZAuthModule/Controller/AccountController.php
```diff
@@ -147,55 +150,88 @@ public function lostPasswordAction(Request $request)
...
+            $this->addFlash('error', $this->__('Your request could not be processed due to invalid arguments.'));
+
+            return $this->redirectToRoute($redirectToRoute);
+        }
+
+        $user = $this->get('zikula_users_module.user_repository')->findBy([
+            'uid' => $requestDetails['userId'],
+            'uname' => $requestDetails['userName'],
+            'email' => $requestDetails['emailAddress']
+        ]);
```
craigh: when using the `uid` you will only get one record or null. There is no possibility of more. you can simply use the `uid` and the `find` method and then check for null.

also this is from the `user_repository`. Anyone that has logged in since upgrade past 1.4.3 will have had their record migrated to the zauth_authentication_mapping table. So this will not work.

craigh:
to review in combination with the #3177: User data is migrated from the `users` table to the `zauth_authentication_mapping` table when each user logs in after that upgrade. This means that *some* user data will be in both places - the only thing this PR cares about - the password - is the main thing that is migrated.

This means you will need to have a 'new' method which uses `zauth_authentication_mapping` and a BC method using the `users` table - the easier it is to remove the BC layer, the better please. 

I didn't walk though each line of this PR yet, but it appears you haven't considered the `zauth_authentication_mapping` location for passwords. perhaps I missed it

src/system/ZAuthModule/Controller/AccountController.php
Guite: well, I did not invent this, but rewrite the code which was there; so if it does not work it didn't work before, too

Guite:
> This means you will need to have a 'new' method which uses zauth_authentication_mapping and a
> BC method using the users table - the easier it is to remove the BC layer, the better please.

> I didn't walk though each line of this PR yet, but it appears you haven't considered the
> zauth_authentication_mapping location for passwords. perhaps I missed it

again: I reused the code which was there; so if it has serious problems those are general and very probably subject of another issue

src/system/ZAuthModule/Controller/AccountController.php
craigh: that's simply not true. The previous code only searched based on email. Such a search may possibly return several rows because in the past, the email has not been required to be unique. You are searching based on multiple params including the uid, which must be unique, therefore it will only return one row and using the `find` method is appropriate.

Guite: I meant

> also this is from the user_repository. Anyone that has logged in since upgrade past 1.4.3 will have had their record migrated to the zauth_authentication_mapping table. So this will not work.

craigh:
> again: I reused the code which was there; so if it has serious problems those are general and very probably subject of another issue

if it works, great! I'm just asking you to check both possibilities: 
  1. a legacy login (user has not logged in since the upgrade to 1.4.3) where the pwd will be stored in the users table 
  2. a new login (user was created after 1.4.3 install or has logged in since) where the pwd will be store in the zauth table.

src/system/ZAuthModule/Controller/AccountController.php
craigh: please [comment below](https://github.com/zikula/core/pull/3178#issuecomment-260350099)

Guite:
Tested both. Both work.

Issue #3171, opened 2016-11-12
cant get category based permissions to work


MrMontesa:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.3
| PHP Version       | 5.6.25

### Expected behavior
Trying to hide News with a certain category from being displayed to not registered visitors.
Or: supress displaying Postcalendar events based on category. unregistered users should not have access to certain events which belong to the same category.

### Actual behavior
News or Postcalendar events are still shown.

### Steps to reproduce
Create News category
add permission rule that surpresses this news from being displayed.
Permission help suggests: ZikulaCategoriesModule::Category      ID::
Guite advised the following from german forum: ZikulaCategoriesModule:Main:Category ::

Both doesnt seem to work.
Let me know if I can be of any help making this work

Guite:
First thing to note here is that both News and PostCalendar use the `CategoryUtil::hasCategoryAccess()` method for checking if access to a certain category is given or not.

https://github.com/zikula-modules/News/blob/master/lib/News/ResultChecker.php#L24
https://github.com/craigh/PostCalendar/blob/master/lib/PostCalendar/Api/Event.php#L128
https://github.com/craigh/PostCalendar/blob/master/lib/PostCalendar/Api/Event.php#L175

Inside this method the actual permission check is done here: https://github.com/zikula/core/blob/1.4/src/lib/util/CategoryUtil.php#L1663

Guite:
Can you check whether `ZikulaCategoriesModule:Main:Category .*` makes a difference (note the instance `.*` instead of `::`)?

MrMontesa:
Not sure if I got your suggestion correctly. I've set Component to "ZikulaCategoriesModule:Main:Category" and Instance to "10014.*" which is a Postcalendar category ID. No change in behavior.

AS a side note the "info about permissions" popup would need an update to list ZikulaCategoriesModule:Main:Category. Right now it's without :Main:.

Guite:
The component looks correct. Please try the following:

1. Change the instance to `10014::`.
2. Does it work then?
3. If not, do the changes shown in #3189
4. Does it work now?

MrMontesa:
I had 1. configured before. And I guess the . (dot) at the end of the line is not meant to be used
But cross checked and does not work.
Then I applied the patch for the two files and doesnt work either.
Deleted cache, but no difference.

PR #3189, opened 2016-11-15
[WIP] simplify category permission scheme

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3171 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no


craigh:
wouldn't this be a BC break?

Guite:
If something does not work or is not reasonable it is already kind of broken, isn't it?
And how reasonable is it to require potentially long pathes in a permission instance?

Beside this it was just an attempt to *find* the problem.

Issue #2089, opened 2014-12-07
Double email to admin on user registration


craigh:
Wrong-ish admin mails on new registration

you as an admin get the same email twice: Once when the user registers, second when he verifys his email.

cmfcmf:
Okay, I've just reproduced it:
1. Activate "moderate registrations"
2. Insert an "E-mail address to notify of registrations"
3. Register a new account -> both admin and user get an email
4. verify account using the link in the user's email
5. approve account using the admin interface
6. both user and admin receive an email saying that the account has now been activated. The admin shouldn't receive one though.

craigh:
@cmfcmf will you be fixing this? If not, please move it to 1.4.1 or 2.0.0

cmfcmf:
Moved to 1.4.1.

rallek:
mmh, for me it sounds like a feature :-)

On one of my sites I do have several unexperienced users. They asked me: 'Hey, I registered at the site, but I still do not have access.' And I can say: 'Yes you registered but still you did not approve'

If I would be asked I would want to stay with that feature. But I did not checked right now if the emails are different.

Guite:
This problem does not exist anymore. I found some other smaller issues in the approval process though (mainly mail-related). Going to provide a PR for these.

PR #3191, opened 2016-11-17
various minor fixes for verification and approval emails, refs #2089

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | #2089 
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/system/UsersModule/Controller/UserAdministrationController.php
+++ b/src/system/UsersModule/Controller/UserAdministrationController.php
@@ -193,6 +193,7 @@ class UserAdministrationController extends AbstractController
         if (!$this->hasPermission('ZikulaUsersModule', '::', ACCESS_MODERATE)) {
             throw new AccessDeniedException();
         }
+
         $forceVerification = $this->hasPermission('ZikulaUsersModule', '::', ACCESS_ADMIN) && $force;
         $form = $this->createForm('Zikula\UsersModule\Form\RegistrationType\ApproveRegistrationConfirmationType', [
             'user' => $user->getUid(),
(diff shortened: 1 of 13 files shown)


src/system/UsersModule/Controller/UserAdministrationController.php
```diff
@@ -201,40 +202,42 @@ public function approveAction(Request $request, UserEntity $user, $force = false
...
 
                 if ($notificationErrors) {
-                    foreach ($notificationErrors as $error) {
-                        $this->addFlash('error', $error);
-                    }
-                } else {
-                    $this->addFlash('status', $this->__f('Done! %sub% has been approved.', ['%sub%' => $user->getUname()]));
+                    $this->addFlash('error', implode('<br />', $notificationErrors));
                 }
+                $this->addFlash('status', $this->__f('Done! %sub% has been approved.', ['%sub%' => $user->getUname()]));
```
craigh: won't this say `approved` every time? even if there were errors?

src/system/UsersModule/Helper/MailHelper.php
```diff
@@ -101,11 +110,14 @@ public function createAndSendRegistrationMail(UserEntity $userEntity, $userNotif
         }
         if ($adminNotification) {
             // mail notify email to inform admin about registration
-            $mailEmail = $this->variableApi->get('ZikulaUsersModule', UsersConstant::MODVAR_REGISTRATION_ADMIN_NOTIFICATION_EMAIL, '');
-            if (!empty($mailEmail)) {
-                $mailSent = $this->sendNotification($mailEmail, 'regadminnotify', $rendererArgs);
+            $notificationEmail = $this->variableApi->get('ZikulaUsersModule', UsersConstant::MODVAR_REGISTRATION_ADMIN_NOTIFICATION_EMAIL, '');
+            if (!empty($notificationEmail)) {
+                $authMapping = $this->authenticationMappingRepository->getByZikulaId($userEntity->getUid());
```
craigh: not sure about this. Every user will not have an entry in this table. for example, if a user registers with the OAuth module, they will not have an entry. The UsersModule will have a uname and email stored for each user regardless of registration/login method.

src/system/UsersModule/Controller/UserAdministrationController.php
craigh: of course 👍

src/system/UsersModule/Helper/MailHelper.php
```diff
@@ -113,7 +113,7 @@ public function createAndSendRegistrationMail(UserEntity $userEntity, $userNotif
             $notificationEmail = $this->variableApi->get('ZikulaUsersModule', UsersConstant::MODVAR_REGISTRATION_ADMIN_NOTIFICATION_EMAIL, '');
             if (!empty($notificationEmail)) {
                 $authMapping = $this->authenticationMappingRepository->getByZikulaId($userEntity->getUid());
-                $rendererArgs['isVerified'] = $authMapping->isVerifiedEmail();
+                $rendererArgs['isVerified'] = null === $authMapping || $authMapping->isVerifiedEmail();
```
Guite: This doesn't get the email, but the information about whether the email has been verified. This is not part of the user entity anymore.

craigh: sorry, I'm not reading the code carefully enough. hard for me to multitask. You're right - email verification is part of ZAuth.

craigh: I guess my only remaining concern about this is that I was trying to separate the users and ZAuth as much as possible with the idea that ZAuth could even possibly be disabled or removed. That may still be impossible with or without this PR but this ties them together more, not less.

craigh: merge it then I guess.

Issue #3027, opened 2016-08-23
Upgrade to Symfony 3


Guite:
https://www.cloudways.com/blog/upgrade-symfony-2-to-3/
https://github.com/symfony/symfony/blob/master/UPGRADE-3.0.md

Guite:
Includes #2099

Guite:
Done in #3196

Issue #2639, opened 2015-11-08
ContainerAware class deprecated in Symfony


craigh:
https://github.com/symfony/symfony/commit/807ebac08ce0c7d0e0f9d2a69dc73c7f3c934da8

investigate how this affects Zikula.

a quick search shows that the class in only used once directly:

https://github.com/zikula/core/blob/1.4/src/lib/util/ModUtil.php#L1037

but instances of ContainerAwareInterface are used elsewhere. It is not clear to me how the new `ContainerAwareTrait` should be properly used yet. Also it is not clear to me how to determine if a class requires the injection of the Container on instantiation (e.g. `MyClass->setContainer($c)`) perhaps classes employing `ContainerAwareTrait` will always implement `ContainerAwareInterface`? It is unclear.

craigh:
refs https://www.drupal.org/node/2229183
and https://github.com/symfony/symfony/pull/16411

craigh:
Traits require php > 5.4 so this issue cannot be backported to Core-1.x. It can only be implemented in Core-2.0

craigh:
Core requires php 5.4+ now. so this can be done any time

Guite:
This was already done in `master` as part of #3196

craigh:
if done in `master` then moving to different milestone.

Issue #2099, opened 2014-12-10
Consider switching to the "new" Symfony 3 directory structure


cmfcmf:
http://stackoverflow.com/questions/23993295/what-is-the-new-symfony-3-directory-structure
It _might_ be worth doing so to avoid problems later.

_I haven't fully understood why there is a Symfony 3 directory structure which can be used in Symfony 2.x already, but one of you will probably understand._

Guite:
This is done in `master`.

PR #3196, opened 2016-11-18
Updated to Symfony 3.2.1 final

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3027, #2099, #2639
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes


src/lib/Zikula/Bundle/FormExtensionBundle/Form/Type/LocaleType.php
```diff
@@ -29,13 +30,13 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
         ]);
     }
 
-    public function getName()
+    public function getBlockPrefix()
```
craigh: I believe this change was introduced in Symfony 2.8. If that is the case here (or with other things you are changing) then they should be changed only in the Core-1.4 branch and then merged up to Core-2.0. This means that large monolithic changes like this PR (currently) are probably not the best way to handle this transition.

ANYTHING that will work with Symfony 2.8 must be first merged into Core-1.4 and then merged up.

src/lib/Zikula/Core/Forms/ZikulaCsrfProvider.php
```diff
@@ -11,12 +11,12 @@
...
 namespace Zikula\Core\Forms;
 
-use Symfony\Component\Form\Extension\Csrf\CsrfProvider\CsrfProviderInterface;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 
 /**
  * Csrf provider based on zikula csrf system.
  */
-class ZikulaCsrfProvider implements CsrfProviderInterface
+class ZikulaCsrfProvider implements CsrfTokenManagerInterface
```
craigh: I would be curious to know if the `ZikulaCsrfProvider` is even necessary at all. Why aren't we using the native one?

src/lib/util/ModUtil.php
```diff
@@ -11,7 +11,7 @@
 
 use Symfony\Bundle\FrameworkBundle\Controller\ControllerNameParser;
 use Symfony\Bundle\FrameworkBundle\Controller\ControllerResolver;
-use Symfony\Component\DependencyInjection\ContainerAware;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Zikula\Core\Event\GenericEvent;
```
craigh: changes to ModUtil (or any Util class) in the master branch are a waste of time. They are **all** going to be deleted.

src/system/ThemeModule/Resources/config/services.xml
```diff
@@ -23,9 +23,8 @@
         </service>
 
         <service id="zikula_theme_module.theme_entity.repository"
-                 class="%zikula_theme_module.theme_entity.repository.class%"
-                 factory-service="doctrine"
-                 factory-method="getRepository">
+                 class="%zikula_theme_module.theme_entity.repository.class%">
+            <factory service="doctrine" method="getRepository" />
```
craigh: can this be done in Symfony 2.8?

src/lib/Zikula/Core/Forms/ZikulaCsrfProvider.php
Guite: If we switch to the native one the custom implementation should be removed from 1.4 and 2.0-spec then. But as part of a separate issue.

src/lib/Zikula/Bundle/FormExtensionBundle/Form/Type/LocaleType.php
Guite: See #3202

src/system/ThemeModule/Resources/config/services.xml
Guite: See #3202

Issue #3194, opened 2016-11-18
cannot install Postcalendar on 1.4.4-dev


MrMontesa:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.4-dev
| PHP Version       | 5.6.26

### Expected behavior
Install of Postcalendar extension should work. Since there's not much change in Postcal module but core, I expect this to be a core issue.

### Actual behavior
Once extension folder is placed in modules folder, Zikula spits an error

```

Stack Trace
in system/ExtensionsModule/Helper/ComposerValidationHelper.php at line 142   -
        if (!$validator->isValid()) {
            foreach ($validator->getErrors() as $errorDetails) {
                $this->errors[] = $this->__f('Error found in composer file of %component% (%filePath%) in property "%property%": %error%.', [
                    '%component%' => $this->bundleName,
                    '%filePath%' => $this->filePath,
                    '%property%' => $errorDetails['property'],
```

### Steps to reproduce
Install zk1.4.4-dev
download & extract Postcalendar
access /extensions/module/list page

Guite:
Hi @MrMontesa 

This issue is fixed by #3195. You can apply this on your page. After you cleared your cache, the error is gone.

MrMontesa:
Thanks for the swift turnaround. I can confirm Postcalendar installs now.
But I do receive a new error message after completing the install, which is:

`Error found in composer file of PostCalendar (PostCalendar) in property "extra": The property extra is required.`

Is this PostCal or Core? You want me to file a new ticket for this?
thanks

Guite:
This is a PostCalendar issue. It's composer file is incomplete.

craigh:
I disagree that this is a PC issue. The core should be more flexible with legacy modules I think. I was worried about this when the composer validator was enabled. I wonder if we shouldn't simply disable the validator for non-bundle modules.

craigh:
I am now able to install PostCalendar without error. Please update this ticket if you continue with the same issue.

PR #3204, opened 2016-11-22
update composerValidator

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | #3194
| License           | MIT
| Changelog updated | no

## Description
add tests for composerValidator.
move where validation takes place so that it only validates Bundles.


craigh:
@Guite I think this should work better. Let me know what you think.

Guite:
:+1: 

https://github.com/zikula/core/blob/1.4/src/system/ThemeModule/Helper/BundleSyncHelper.php needs to be updated, too.

PR #3276, opened 2016-12-06
avoid site break if capabililty routes are invalid, fixed #2905

### LINKED ISSUE #2905: removing the default controller/method routes for a module triggers exception
| Q | A |
| --- | --- |
| Zikula Version | 1.4.1+ |
| PHP Version | 5.x |
### Expected behavior

upgrading/developing modules function as expected without breaking site
### Actual behavior

site breaks. at least for admin.
### Steps to reproduce

delete the controller/method of the "user" or "admin" route set in the `capabilities` array in the `composer.json` file and the site breaks.

This is because the information is 'cached' in the DB and not refreshed unless the ExtensionsModule main page is visited. This is an issue because the information is used in the creation of the hidden admin _mmenu_. and so the site throws an exception of 'route not found'.

Perhaps the creator could wrap the route generation in a try/catch block and deal with routes not found some other way.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #2905 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Core/LinkContainer/LinkContainerCollector.php
+++ b/src/lib/Zikula/Core/LinkContainer/LinkContainerCollector.php
@@ -12,6 +12,7 @@
 namespace Zikula\Core\LinkContainer;
 
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Zikula\Core\Event\GenericEvent;
 
 class LinkContainerCollector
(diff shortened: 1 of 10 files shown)


Guite:
This looks like this:

screenshot_20161206_142906

craigh:
maybe add an :warning:

Guite:
done

PR #3277, opened 2016-12-06
search also for user entities during lost user name and lost password processes, fixed #3274

### LINKED ISSUE #3274: User unknown in password recovery (zauth)
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.4 RC5
| PHP Version       | 5.5.30

### Expected behavior
password recovery is working

### Actual behavior
The user is not know anymore if he not logged in one time after the zauth is introduced (after upgrade the core).

### Steps to reproduce
Upgrade from 1.4.1 to 1.4.4 and try to recover password of an already registered user.

#3256 is not solved completely. It is only solved 70%. But this is not a blocker for RC5

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3274
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/system/ZAuthModule/Controller/AccountController.php
+++ b/src/system/ZAuthModule/Controller/AccountController.php
@@ -48,15 +48,28 @@ class AccountController extends AbstractController
         $form->handleRequest($request);
         if ($form->isSubmitted()) {
             $data = $form->getData();
-            $mapping = $this->get('zikula_zauth_module.authentication_mapping_repository')->findBy(['email' => $data['email']]);
+
+            $email = $data['email'];
+            $userName = '';
+
+            $mapping = $this->get('zikula_zauth_module.authentication_mapping_repository')->findBy(['email' => $email]);
             if (count($mapping) == 1) {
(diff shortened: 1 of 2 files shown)


craigh:
I wonder if some of this is really BC code... maybe it could be marked deprecated. This is all mainly because of the transition right?

I intentionally didn't migrate all the users from the `users` table to the `zauth_authentication_mapping` table to minimize upgrade execution time and in order to not have to deal with duplicates, etc. It can be a real headache. The solution was allow users to upgrade themselves as they login. Overtime all the users would be converted. Maybe we can add some kind of CLI command or something to complete the job and then remove code like this in 2.0.

Guite:
> Maybe we can add some kind of CLI command or something to complete the job and then remove code like this in 2.0.

Won't help users without CLI access.
But we could consider an action within the admin area which converts 10 users and reloads itself until the job is done.

PR #3305, opened 2016-12-17
fixed logical operators

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/system/BlocksModule/Controller/BlockController.php
+++ b/src/system/BlocksModule/Controller/BlockController.php
@@ -110,7 +110,7 @@ class BlockController extends AbstractController
             $renderedPropertiesForm = $this->formContentModify($request, $blockEntity);
         }
 
-        if ($form->isSubmitted() and $form->get('save')->isClicked() and $form->isValid()) {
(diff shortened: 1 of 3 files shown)


src/system/BlocksModule/Controller/BlockController.php
```diff
@@ -110,7 +110,7 @@ public function editAction(Request $request, BlockEntity $blockEntity = null)
             $renderedPropertiesForm = $this->formContentModify($request, $blockEntity);
         }
 
-        if ($form->isSubmitted() and $form->get('save')->isClicked() and $form->isValid()) {
+        if ($form->isSubmitted() && $form->get('save')->isClicked() && $form->isValid()) {
```
craigh: `and` is valid http://php.net/manual/en/language.operators.logical.php

Guite: From SensioLabsInsights:

The `and` operator does not have the same precedence as `&&`. This could lead to unexpected behavior, use `&&` instead.

Issue #2601, opened 2015-09-19
jsgettext FC


craigh:
need to figure out how to deal with `{pageaddvar name="jsgettext" value="module_news_js:News"}` going forward.
Not sure it ever properly worked in the first place.
translation for >=1.4 is different anyway and JS translation must be accommodated.
refs https://github.com/zikula/core/issues/2487

currently `pageAddAsset()` doesn't allow `jsgettext` so it is filtered out and set as a pagevar instead. Obviously this is wrong, but the system won't work in twig anyway, so it probably doesn't matter until a solution is implemented in general.

@phaidon

craigh:
maybe a duplicate of #2487

Kaik:
What do you think about this?
https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/doc/index.md

craigh:
yes that is the first choice in the related ticket :wink:  (#2487)

Issue #2487, opened 2015-06-07
Options for JS Translation


craigh:
**This one seems to be the most widely used**
https://github.com/willdurand/BazingaJsTranslationBundle

http://knpbundles.com/funddy/jstranslations-bundle

http://loige.co/integrating-twig-js-and-bazingajstranslationbundle/

PR #3338, opened 2016-12-29
Js translation

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #2601, #2487
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description
Add BazingaJsTranslationBundle and add zikula type translation with tests
## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -46,6 +46,7 @@ CHANGELOG - ZIKULA 1.4.x
     - sensio/generator-bundle updated from 3.0.11 to 3.1.2
     - symfony updated from 2.8.14 to 2.8.15
     - symfony/swiftmailer-bundle updated from 2.3.11 to 2.4.2
+    - swiftmailer updated from 5.4.3 to 5.4.5
     - twig updated from 1.28.2 to 1.30.0
     - twig-extensions updated from 1.3.0 to 1.4.1
     - typeahead.js updated from 1.0.1 to 1.1.0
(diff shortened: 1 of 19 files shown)


craigh:
I'm not sure this is complete, but it is worth considering for 1.4.5 even as is.

craigh:
this includes the swift mailer update

craigh:
One thing I don't like is that this method sets the arguments without the '%' signs. And so this differs from the php method.

Guite:
This is awesome :+1: 

Please add an example to the Markdown file for how a module should implement this: How must the catalogue be named? Does it need to be initialised somehow? Any JS files to include? Etc.

Guite:
> how a module should implement

module or theme of course :-)

Guite:
Also the following files contain (commented) calls of `Zikula.__` which might be converted:

- system/CategoriesModule/Resources/public/js/categories_admin_edit.js
- system/CategoriesModule/Resources/public/js/categories_admin_view.js
- system/PageLockModule/Resources/public/js/PageLock.js
- system/RoutesModule/Resources/public/js/ZikulaRoutesModule.Validation.js
- system/ZAuthModule/Resources/public/js/Zikula.ZAuth.PassMeter.js

Guite:
> Also the following files contain (commented) calls of Zikula.__ which might be converted:

Did that in https://github.com/zikula/core/pull/3338/commits/902240ec25cd96ef4a6609a1997f50fee2c6d104

craigh:
> This is awesome

thanks!

> Please add an example ...

> I'm not sure this is complete

😆

Issue #3388, opened 2017-01-16
Implement Composer Merge plugin


craigh:
https://github.com/wikimedia/composer-merge-plugin

@cmfcmf @Guite 

this could be an answer to dynamically including the `composer.json` files from included bundles/modules/themes.

e.g.

            "include": [
                "composer.local.json",
                "system/*/composer.json"
                "modules/*/composer.json"
                "themes/*/composer.json"
            ],

craigh:
another idea would be to fire an event in the `getBundles()` method of the kernel (is the event dispatcher available then?) to get additional bundles that must be included in the container.

craigh:
since the container is not available at that point, I'm guessing the event dispatcher is not. Another solution would be to add "bundle dependencies" to the composer "extra" field. Parse that and add the dependencies to the kernel.

craigh:
Drupal related issues
 - https://www.drupal.org/node/2444459
 - https://www.drupal.org/node/2609568
 - https://github.com/paul-m/drupal-merge-plugin
 - https://bojanz.wordpress.com/2015/09/18/d8-composer-definitive-intro/
 - https://glamanate.com/blog/managing-your-drupal-project-composer
 - http://drupal.stackexchange.com/questions/203525/managing-module-library-dependencies-with-composer

Guite:
Is this about automatically including vendors required in modules/themes?

craigh:
yes.

it could solve a great many problems for us in terms of vendor dependencies, but it could also solve other problems with module dependencies and even allow us to further modularize the core and build it simply with composer. 

it also raises as many questions as it solves. Some that I can think of: this will require running `composer update` on module install/upgrade/remove. this is a **very** long process and not easily done in a browser. Is this possible? Is there a way to avoid it (for example if the module has no unresolved dependencies). Will we have to ship composer with zikula then? (seems so).

Guite:
Possibly related:
https://github.com/dflydev/dflydev-embedded-composer
https://github.com/shadiakiki1986/composer-wrapper

craigh:
this could also allow quickly customized downloads with preconfigured modules installed based on a generated composer.json file. Could easily lead to several options for a "CMS" package for example.

cmfcmf:
This was also related: https://github.com/zikula/core/issues/909#issuecomment-148917664
PageKit is already using Composer to install dependencies on the fly, IIRC.

Guite:
After having read a bit about this topic I think we need to see two concerns separately:

[...]

PR #3437, opened 2017-02-03
Handle module dependencies using composer

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3388 
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

This utilises the Composer Merge Plugin for initial composer executions.


composer.json
```diff
@@ -131,31 +63,25 @@
...
-            "type": "vcs",
-            "url": "https://github.com/corejavascript/typeahead.js"
+        "merge-plugin": {
+            "include": [
+                "src/lib/Zikula/Bundle/*/composer.json",
+                "src/system/*/composer.json",
+                "src/modules/*/composer.json",
+                "src/modules/*/*/composer.json",
+                "src/themes/*/composer.json",
+                "src/themes/*/*/composer.json"
```
craigh: isn't that redundant? doesn't `*` support multi-level?

Guite: no it does not

craigh: that could be a problem because there is no guarantee of the composer.json placement level

craigh:
I wasn't clear this was ready. if so (and you know it works) then merge it

Issue #3465, opened 2017-02-14
Google Fonts Need to Load First


shefik:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4.6
| PHP Version       | 5.6

### Expected behavior

Google Fonts (or any other third-party font stylesheets) need to load first, before other CSS files, when the CSS assets are combined into one file, since the combined files reference required fonts.

### Actual behavior

The combined CSS file is loaded first, then the Google Fonts stylesheets are loaded afterwards, even though the weight for the Google Fonts stylesheets is set to `0`.

### Steps to reproduce

In your theme add the following:

```
{{ pageAddAsset('stylesheet', '//fonts.googleapis.com/css?family=Montserrat:400,700', 0) }}
{{ pageAddAsset('stylesheet', '//fonts.googleapis.com/css?family=Raleway:100,300,600,800', 0) }}
```

Set your config to combine the CSS asset files. the output will be something like:

```
<link rel="stylesheet" href="/templates/combined_asset/css/1da10173c7b129a09bbcedafd07bfd441186400.css" type="text/css">
<link rel="stylesheet" href="//fonts.googleapis.com/css?family=Raleway:100,300,600,800" type="text/css">
<link rel="stylesheet" href="//fonts.googleapis.com/css?family=Montserrat:400,700" type="text/css">
```

craigh:
of course, the faster answer here would have been to manually use a 'normal' reference instead of using the zikula method, but I've accommodated the request. I've also added more docs

PR #3466, opened 2017-02-14
Switching order of array, to load remote CSS file first

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | https://github.com/zikula/core/issues/3465
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description

Switching order of array, to load remote CSS file first, when merging.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -16,7 +16,8 @@ CHANGELOG - ZIKULA 1.4.x
     - Added Permission-based controls for MenuModule menu items (#3314).
 
  - Vendor updates:
-    - ?
+    - swiftmailer/swiftmailer updated from v5.4.5 to v5.4.6 
+    - willdurand/js-translation-bundle updated from 2.6.3 to 2.6.4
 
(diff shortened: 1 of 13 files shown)


craigh:
I don't know the circumstances in which someone might wish to include any external asset. Couldn't this _cause_ as many problems as it solves? for example what if the external asset was dependent on jQuery (which is loaded later)?

shefik:
This is only switching the load of CSS, not JavaScript. So, jQuery would not be affected.

craigh:
no. It affects both. please review the code.

shefik:
Okay, I only meant to change loading for CSS, not JS.

craigh:
> Okay, I only meant to change loading for CSS, not JS.
OK. update the PR or close it then as this will affect both.

Guite:
ping @shefik

PR #3486, opened 2017-03-09
added automatical polyfill initialisation, fixed wrong service argument, fixed #3348

### LINKED ISSUE #3348: Use polyfills in all forms
See https://github.com/zikula/core/issues/3342#issuecomment-269862441

Either include the Twig method call in all templates or find a way to call it centrally for all forms.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3348
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

### DIFF
--- a/src/javascript/polyfiller.init.js
+++ b/src/javascript/polyfiller.init.js
@@ -5,5 +5,5 @@
         method: 'GET'
     });
     $.webshims.activeLang(Zikula.Config.lang);
-    $.webshims.polyfill(typeof Zikula.Config.polyfillFeatures !== 'undefined' ? Zikula.Config.polyfillFeatures : []);
+    $.webshims.polyfill(Zikula.Config.polyfillFeatures || []);
 })(jQuery);
--- a/src/javascript/polyfiller.init.js
+++ b/src/javascript/polyfiller.init.js
@@ -5,5 +5,5 @@
         method: 'GET'
     });
     $.webshims.activeLang(Zikula.Config.lang);
-    $.webshims.polyfill(Zikula.Config.polyfillFeatures);
+    $.webshims.polyfill(typeof Zikula.Config.polyfillFeatures !== 'undefined' ? Zikula.Config.polyfillFeatures : []);
 })(jQuery);
--- a/src/lib/Zikula/Bundle/FormExtensionBundle/Listener/FormDetectionListener.php
+++ b/src/lib/Zikula/Bundle/FormExtensionBundle/Listener/FormDetectionListener.php
@@ -12,7 +12,10 @@
 namespace Zikula\Bundle\FormExtensionBundle\Listener;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\StreamedResponse;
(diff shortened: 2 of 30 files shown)


src/system/ThemeModule/Resources/config/theme_engine_event_listeners.yml
```diff
@@ -60,7 +60,7 @@ services:
           - "@zikula_extensions_module.api.variable"
           - "@zikula_users_module.current_user"
           - "@templating.engine.twig"
-          - "@zikula_core.common.theme.themevars"
+          - "@zikula_core.common.theme.pagevars"
```
Guite: This fixes a bug which caused that polyfills never went into the `ZConfig` array.

src/lib/Zikula/Bundle/FormExtensionBundle/Listener/FormDetectionListener.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Listener for the `kernel.response` event.
+     *
+     * @param FilterResponseEvent $event The event instance
+     */
+    public function onResponse(FilterResponseEvent $event)
+    {
+        $response = $event->getResponse();
+        $content = $response->getContent();
```
stof: you should detected BinaryFileResponse and StreamedResponse before that to avoid calling ``getContent`` on them (which return ``false``), and also ``JsonResponse`` (to avoid a useless check).
Btw, you should probably also check the content type, to inject this only in HTML responses

src/javascript/polyfiller.init.js
```diff
@@ -5,5 +5,5 @@
         method: 'GET'
     });
     $.webshims.activeLang(Zikula.Config.lang);
-    $.webshims.polyfill(Zikula.Config.polyfillFeatures);
+    $.webshims.polyfill(typeof Zikula.Config.polyfillFeatures !== 'undefined' ? Zikula.Config.polyfillFeatures : []);
```
stof: ``Zikula.Config.polyfillFeatures || []`` may be enough here (and shorter)

Guite: Thanks, changed that.

src/lib/Zikula/Bundle/FormExtensionBundle/Listener/FormDetectionListener.php
Guite: Thank you @stof, I've added checks for unwanted response types.

src/lib/Zikula/Bundle/FormExtensionBundle/Twig/Extension/FormExtension.php
```diff
@@ -34,20 +54,23 @@ public function __construct($container = null)
...
         ];
     }
 
-    public function polyfill(array $features = ['forms'])
+    /**
+     * Adds polyfill features to be included into the page.
+     *
+     * @param array $features List of desired polyfills
+     */
+    public function polyfill(array $features = ['forms', 'forms-ext'])
```
craigh: I wonder why github is tagging that with red

Guite: no idea

stof: because the PHP syntax highlighter they use does not like the short array syntax in default values (it does not like default values for arguments with a scalar typehint either btw).
This might be a contribution idea for someone wanting to improve syntax highlighting btw.

PR #3513, opened 2017-04-01
moved PageLock to dedicated repository, fixed #3283

### LINKED ISSUE #3283: Consider outsourcing PageLock
| Q                 | A
| ----------------- | ---
| Zikula Version    | 2.0
| PHP Version       | n/a

I think we should consider outsourcing PageLock for Core-2.0

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3283 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no


Guite:
@craigh There is maybe some update required for the meta data stored in the database (`modules` table). Any idea where to place this ideally?
Or would it be updated automatically during bundle sync when visiting the extensions list?

craigh:
> Or would it be updated automatically during bundle sync when visiting the extensions list?

I think it will be- but also need to write a 1.5 upgrade routine in the CoreInstallerBundle - this can be done separately from this PR

src/lib/Zikula/Bundle/CoreBundle/composer.json
```diff
@@ -17,6 +17,7 @@
 
         "zikula/legal-module": "dev-master",
         "zikula/profile-module": "dev-master",
+        "zikula/pagelock-module": "dev-master",
```
craigh: change to 1.2.0

Guite: changed to `1.2.*`

PR #3520, opened 2017-04-05
Update JCSSUtil.php

gfr:
Added missing web to the asset path.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
Added missing web to bootstrap path and changed polyfiller.init.js path in legacy JCSSUtil.php

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog


gfr:
Unfortunately I have not tried it with 1.4.6 but I think so

craigh:
it would make sense if this was 1.5/master only because I just moved all these assets a few days ago. Also, I was developing in the master branch, so the legacy paths would be incorrect. can you check and see if any other assets in JCSSUtil are affected?

Issue #3454, opened 2017-02-09
second cookie for ajax?


craigh:
refs 
 - https://github.com/zikula/core/commit/9f1b4489cf264605f9c73a0ff8ee85cbc55b2519
 - https://github.com/zikula/core/issues/3450#issuecomment-278561864
 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies

is there a way to make the session cookie http only and create a second (non http only) cookie that could be used by ajax thus making the session cookie more secure?

craigh:
This is related to https://github.com/zikula/core/issues/3206

Guite:
@craigh as this is a FR: move to 1.5.0 or 2.1.0?

craigh:
https://gist.github.com/Burgov/7300028

craigh:
Files involved:
 - ThemeModule/EventListener/DefaultPageAssetSetterListener.php
 - ThemeModule/EventListener/AddJSConfigListener.php
 - ThemeModule/Resources/views/Engine/JSConfig.html.twig
 - Bundle/CoreBundle/Resources/public/js/jquery_config.js

Guite:
@craigh if you think this is important please move back to 1.5

craigh:
refs `\Zikula\Bundle\HookBundle\Controller\HookController::checkAjaxToken`

Guite:
refs `cookie_httponly` see https://github.com/zikula/core/blob/master/src/app/config/config.yml#L31

Guite:
Fixed by #3895

Other related tasks are covered by #3206 and #3505

Issue #2186, opened 2015-01-18
Enable CSRF protection in install/upgrade forms


craigh:
CSRF form protection is currently disabled in all CoreInstallerBundle forms (see for example:  https://github.com/zikula/core/blob/1.4/src/lib/Zikula/Bundle/CoreInstallerBundle/Form/Type/CreateAdminType.php#L74)

It needs to be re-enabled when it can be figured out why it doesn't work. I'm guessing some kind of session issue.

refs #2172 #2187

Guite:
This is definitely related to sessions, but very difficult to find.

The token is set in `SessionTokenStorage#setToken()`:

    $this->session->set($this->namespace.'/'.$tokenId, (string) $token);
    die('AAAA: ' . print_r($this->session->all()));

In the next request it is checked in `SessionTokenStorage#hasToken()`:

        die('AAAA: ' . print_r($this->session->all()));
        return $this->session->has($this->namespace.'/'.$tokenId);

In the second `die()` statement the array is empty.

Guite:
Moved to 1.4.6.

Guite:
See also https://github.com/symfony/symfony/issues/13464

PR #3522, opened 2017-04-05
allow sessions to file. fixes #2001, #2186 and refs #3454

### LINKED ISSUE #2001: Storing Sessions in File Not Working
When a site is configured to store sessions in file (instead of a database), the sessions do not get get written to the file.

/index.php?module=security-center&type=admin&func=modifyconfig

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | fixes #2001, #2186 
| Refs tickets      | #3454
| License           | MIT
| Changelog updated | yes

### DIFF
--- a/src/docs/Core-2.0/Session.md
+++ b/src/docs/Core-2.0/Session.md
@@ -1,0 +1,17 @@
+Session Information
+===================
+
+Sessions can be stored in the Database or in the Filesystem. This choice as well as a few configuration options
+can be set in the ZikulaSecurityModule:
+
+The following configuration options are converted to a dynamic config and stored in `/src/app/config/dynamic/generated.yml`
+ - zikula.session.name: _zsid
+ - zikula.session.handler_id: 
+    'session.handler.native_file' #default 
+    or 
+    'zikula_core.bridge.http_foundation.doctrine_session_handler'
+ - zikula.session.storage_id: 
+    'zikula_core.bridge.http_foundation.zikula_session_storage_doctrine' #default 
+    or 
+    'zikula_core.bridge.http_foundation.zikula_session_storage_file'
+ - zikula.session.save_path: '%kernel.cache_dir%/sessions' #default symfony value
--- a/src/docs/Core-2.0/Session.md
+++ b/src/docs/Core-2.0/Session.md
@@ -2,7 +2,7 @@ Session Information
 ===================
 
 Sessions can be stored in the Database or in the Filesystem. This choice as well as a few configuration options
-can be set in the ZikulaSecurityModule:
+can be set in the ZikulaSecurityCenterModule:
 
 The following configuration options are converted to a dynamic config and stored in `/src/app/config/dynamic/generated.yml`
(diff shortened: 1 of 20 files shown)


craigh:
hmm - a problem with the CLI invalid csrf ... 🤔 😠 

@matthiasnoback - do you know why symfony-console-form is using the tokens? I thought it would ignore them? see https://travis-ci.org/zikula/core/builds/219052481

craigh:
well I canme up with a sort of hacky fix to the csrf problem when using cli.

src/docs/Core-2.0/Session.md
```diff
@@ -0,0 +1,17 @@
+Session Information
+===================
+
+Sessions can be stored in the Database or in the Filesystem. This choice as well as a few configuration options
+can be set in the ZikulaSecurityModule:
```
Guite: Typo. How about `Zikula SecurityCenter module`?

src/lib/Zikula/Bundle/CoreInstallerBundle/Stage/Install/CompleteStage.php
```diff
@@ -60,14 +60,19 @@ public function getTemplateParams()
 
     public function getResponse(Request $request)
     {
+        $router = $this->container->get('router');
         if ($this->sendEmailToAdmin($request)) {
             $request->getSession()->getFlashBag()->add('success', $this->__('Congratulations! Zikula has been successfully installed.'));
+            $request->getSession()->getFlashBag()->add('info', $this->__f(
+                'Session are currently configured to use the filesystem. It is recommended that you change this to use the database. Click %here% to configure.',
+                ['%here%' => '<a href="' . $router->generate('zikulasecuritycentermodule_config_config') . '">' . $this->__('Security Center') . '</a>']
+            ));
```
Guite: Would it be possible to change it to the database automatically during/after the `CompleteStage`?

matthiasnoback:
@craigh Indeed, I thought the `csrf_protection` was always set to `false` when using SymfonyConsoleForm. Maybe a newer form version has broken it? Don't know.

src/lib/Zikula/Bundle/CoreInstallerBundle/Stage/Install/CompleteStage.php
craigh: it forces a re-login, so I didn't think that was the best method.

craigh:
@matthiasnoback - we are using an older version of your library with this branch - could that be the cause? Was there anything special with the form classes I was to do?

matthiasnoback:
@craigh Honestly, don't know.

craigh:
@matthiasnoback looks like I just needed to update the library. I was using an older version (v1) because this branch originally was restricted to php >= 5.4, but now it is updated to require php >=5.5 and so I can use v2 of the lib and this fixes the problem.

PR #3537, opened 2017-04-09
Fixed pixel display bug in the Bootstrap theme when the admin panel i…

paustian:
…s pulled out.

| Q                 | A
| ----------------- | ---
| Bug fix?          | [yes]
| New feature?      | [no]
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | 3449
| Refs tickets      | -
| License           | MIT
| Changelog updated | [yes?]

## Description
Found the css bug that was making the Bootstrap theme display poorly when the Admin Interface was pulled out.
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/src/system/AdminModule/Resources/public/css/mmenu-hiddenpanel-customwidth.css
+++ b/src/system/AdminModule/Resources/public/css/mmenu-hiddenpanel-customwidth.css
@@ -1,7 +1,6 @@
 .mm-menu.mm-hiddenpanel-customwidth {
   width: 80%;
-  min-width: 325px;
-  max-width: 325px; }
+  min-width: 325px; }
 
 html.mm-opening.mm-hiddenpanel-customwidth .mm-slideout {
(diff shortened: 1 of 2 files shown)


craigh:
hmm - I think that body padding was put there for a reason. 😄 I wonder why? Did you have any modules installed while testing? Try installing Pages and view the demo page in each theme.

craigh:
**Before changes** in this PR:
screen shot 2017-04-09 at 7 32 28 am

**After**
screen shot 2017-04-09 at 7 31 30 am

So I think we need a different solution 😉

craigh:
also - we should chat about your process again. you did not create a 'topic branch' which is something you want to do. This will make it easier to keep your local branches up to date. Catch me on slack and we can discuss how to do that.

craigh:
@paustian

PR #3547, opened 2017-04-14
Twig extensions improvement

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
simplify twig extensions. move some function to better locations. reduce dependency injection where possible. improve testability (when they get written)


src/lib/Zikula/Bundle/CoreBundle/Tests/Translation/TwigFileExtractorTest.php
```diff
@@ -132,15 +137,20 @@ private function extract($file, ZikulaTwigFileExtractor $extractor = null)
...
+            ->disableOriginalConstructor()
+            ->getMock();
 
         $env = new \Twig_Environment();
         $env->addExtension(new SymfonyTranslationExtension($translator = new IdentityTranslator(new MessageSelector())));
         $env->addExtension(new TranslationExtension($translator, true));
         $env->addExtension(new RoutingExtension(new UrlGenerator(new RouteCollection(), new RequestContext())));
         $env->addExtension(new FormExtension(new TwigRenderer(new TwigRendererEngine())));
-        $env->addExtension(new GettextExtension(new \Zikula\Common\Translator\IdentityTranslator(new MessageSelector()), $kernel));
+        $env->addExtension(new GettextExtension($zikulaTranslator = new ZikulaIdentityTranslator(new MessageSelector()), $kernel));
```
Guite: Please do the assignment of `$translator` in a separate line.

craigh:
https://github.com/zikula/core/blob/fcdcdec106273503ec832e029fe325a617a222d7/src/docs/Twig/Functions.md#L20-L20

`defaultPath` is to generate a url for the default path (user|admin) from composer.json. I don't think it is used anywhere any longer. I think it was a solution to a problem that I since solved differently.

PR #3610, opened 2017-06-21
added user live search form plugin from generator, fixed #710

### LINKED ISSUE #710: Add formplugin for user name including live search
Provide a simple-to-use plugin for templates for user names.
This plugin shoud
1. validate if user exists (regular input field)
2. be extendable (parameter or separate plugin) with an live user search as it is implemented in the Users module for example

It always takes much time to add a live user search into the own templates. Providing a plugin has two effects:
1. every user name input field in any module will look in the same way
2. developers will save much time...

Thanks!

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #710 
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes


craigh:
There are many unrelated (and unnecessary) changes in this PR. There's no sense changing it now, but in the future I'd rather keep a PR focused (if possible) on its intended change - in this case _adding user live search form plugin_. changing all the classnames and 'touching' all these files is a waste of effort. It is done already in master and not worth your time. In this case it makes it difficult to see related/unrelated changes. I know that I will mix things together too sometimes - but I shouldn't ;-)

Issue #3548, opened 2017-04-14
Theme stylesheet should be automatically included


craigh:
currently the front module's stylesheet is automatically included with no additional coding. But Theme stylesheets are required to be [added manually](https://github.com/zikula/core/blob/1.5/src/themes/BootstrapTheme/Resources/views/Include/header.html.twig#L9)

perhaps we should automatically added theme stylesheets also.

shefik:
I agree.

paustian:
I agree too. There are almost always styles you want to include. Is there also a mechanisms to avoid adding it twice? In case someone adds in manually

PR #3627, opened 2017-07-06
Auto theme stylesheets

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3548 
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

### DIFF
--- a/src/lib/Zikula/Core/AbstractBundle.php
+++ b/src/lib/Zikula/Core/AbstractBundle.php
@@ -17,6 +17,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 use Zikula\Bundle\CoreBundle\Bundle\Scanner;
 use Zikula\Bundle\CoreBundle\Bundle\MetaData;
+use Zikula\ThemeModule\Engine\AssetBag;
 
 abstract class AbstractBundle extends Bundle
 {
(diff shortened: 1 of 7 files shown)


src/lib/Zikula/Core/AbstractBundle.php
```diff
@@ -206,6 +207,22 @@ public function getContainer()
...
+     */
+    public function addStylesheet($name = 'style.css')
+    {
+        try {
+            $styleSheet = $this->getContainer()->get('zikula_core.common.theme.asset_helper')->resolve('@' . $this->getName() . ":css/$name");
+        } catch (\InvalidArgumentException $e) {
+            $styleSheet = '';
+        }
+        if (!empty($styleSheet)) {
+            $this->container->get('zikula_core.common.theme.assets_css')->add([$styleSheet => AssetBag::WEIGHT_DEFAULT]);
```
Guite: Better use a higher weight to avoid any other style sheets after it?

Guite: I mean: a higher weight for themes. `WEIGHT_DEFAULT` is still okay for modules.

Guite: Proposal:

- add `const WEIGHT_THEME = 200;`
- add something like `$weight = $this instanceof AbstractTheme ? AssetBag::WEIGHT_THEME : AssetBag::WEIGHT_DEFAULT;`

craigh: you'll notice that in current implementation, theme stylesheets are added at the default weight and this has not been an issue. (e.g. `{{ pageAddAsset('stylesheet', zasset('@ZikulaBootstrapTheme:css/style.css')) }}`)

craigh: also, I think you are mixing up _weight_ and _priority_ (as is used by listeners, etc). 
>  *  - A lighter weight loads before a heavier weight.

Guite: My point was theme stylesheets should be loaded after other ones so that it becomes easier to override and/or customise module layouts.

Issue #3669, opened 2017-07-17
move require.js files to subdir


craigh:
![screen shot 2017-07-17 at 7 44 22 am](https://user-images.githubusercontent.com/350048/28266710-65277540-6ac4-11e7-83dc-ce7336d7aedc.png)

the `require*` files should be in own directory `/require` and the commands that create and call them be adjusted

PR #3671, opened 2017-07-17
attempt to move require.js files into subfolder

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3669 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Composer/CopyRequirementsFile.php
+++ b/src/lib/Zikula/Composer/CopyRequirementsFile.php
@@ -12,8 +12,8 @@
 namespace Zikula\Composer;
 
 use Composer\Script\Event;
-use Symfony\Component\Filesystem\Filesystem;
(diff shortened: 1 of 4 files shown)


craigh:
are these files never required for anything? Why do we have them? @Kaik

craigh:
@Guite this probably requires a changelog as it is a BC break if the path changed

Guite:
> are these files never required for anything?
No, afaik they are there for people who may need them.

> requires a changelog
added in https://github.com/zikula/core/commit/56515a73b803ed8ab587eaf599171b8f54e908e7

Kaik:
Roblach Component Installer is installing it by default we do not use them afaik.

PR #3618, opened 2017-06-29
Hook overhaul

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | yes
| Fixed tickets     | #2785, #2784, #2262, #236
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

PR #3683, opened 2017-07-20
Reenabled support for legacy hooks

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | https://github.com/zikula-modules/Scribite/issues/165
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/lib/Zikula/Bundle/HookBundle/Dispatcher/HookDispatcher.php
+++ b/src/lib/Zikula/Bundle/HookBundle/Dispatcher/HookDispatcher.php
@@ -42,20 +42,39 @@ class HookDispatcher implements HookDispatcherInterface
     private $dispatcher;
 
     /**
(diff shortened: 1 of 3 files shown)


craigh:
This is exactly the code that #3608 and #3618 removed in favor of better solutions. This cannot be re-added because it requires a ContainerBuilder which is not allowed any more. We need a different solution

Guite:
Added #3684 
But honestly I don't see why the old code can't be kept in 1.x.

craigh:
reopened so the branch doesn't get deleted.

I'm looking in to it.

craigh:
@Guite please see https://github.com/zikula/core/commit/1aae6f400a482ac9073902451c7b3c0a2f34c561

PR #3699, opened 2017-07-25
search Profile data; link user results to Profile pages

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | https://github.com/zikula-modules/Profile/issues/82
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

### DIFF
--- a/src/system/SearchModule/Twig/TwigExtension.php
+++ b/src/system/SearchModule/Twig/TwigExtension.php
@@ -129,7 +129,7 @@ class TwigExtension extends \Twig_Extension
     {
         if ($url instanceof ModUrl) { // @deprecated
             return $url->getUrl();
-        } 
+        }
         if ($url instanceof RouteUrl) {
             try {
                 return $this->router->generate($url->getRoute(), $url->getArgs()) . $url->getFragment();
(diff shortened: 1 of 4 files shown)


craigh:
don't you think the profile module should be searching for its own data? why should the user module be responsible for this? My point of https://github.com/zikula-modules/Profile/issues/82 was that Profile should implement the new Searchable api (e.g. `zikula.searchable_module` tag) not that the core should be searching profile data.

Guite:
Actually these are user data (attributes).
Initially I started with a Profile search helper, but this makes no sense, since Users is already searching for users. Profile modules extend users, but they still represent users.

craigh:
Yes, I understand that they re user attributes.

> but this makes no sense, since Users is already searching for users

I disagree. SOC regardless of performance issues, Profile should search for Profile data and Users should search for user (specific) data. Consider another module that might add user attributes. This data may get included in search results. A profile module would know how to limit its own search to only obtain its own data.

Guite:
Proper SOC would require that no user attributes are used at all, but that Profile stores it's data into a dedicated entity.

From a user's point of view it makes no sense if a user appears twice in the search results.

craigh:
another use case what if the profile module wants to control the visibility of its data. If the search is performed in Users, this would be impossible.

PR #3714, opened 2017-07-26
correct mistaken progression in 1.5 branch.

### LINKED ISSUE #3713: undocumented BC break in hooks
see https://github.com/cmfcmf/MediaModule/issues/94

basically, if a module had type hinted `Zikula_HookDispatcher` then it breaks because this class has been replaced with `Zikula\Bundle\HookBundle\Dispatcher\HookDispatcher` and there is no inheritance to maintain BC

this should either be documented or fixed.

One possible fix is outlined in the ticket:
> I think one solution here (in the core) would be to alias `Zikula_HookDispatcher` to `Zikula\Bundle\HookBundle\Dispatcher\HookDispatcher` to maintain BC

I'm not sure if this might make things worse or better. :|

feedback? @cmfcmf @Guite

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description
revert to legacy HookDispatcher class

fixes https://github.com/cmfcmf/MediaModule/issues/94
fixes #3713

### DIFF
--- a/src/lib/Zikula/Bundle/HookBundle/Dispatcher/HookDispatcher.php
+++ b/src/lib/Zikula/Bundle/HookBundle/Dispatcher/HookDispatcher.php
@@ -13,9 +13,6 @@ namespace Zikula\Bundle\HookBundle\Dispatcher;
 
 use Symfony\Component\EventDispatcher\Event;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
-use Zikula\Bundle\HookBundle\Bundle\SubscriberBundle;
-use Zikula\Bundle\HookBundle\Bundle\ProviderBundle;
-use Zikula\Bundle\HookBundle\Collector\HookCollectorInterface;
 use Zikula\Bundle\HookBundle\Dispatcher\Exception\LogicException;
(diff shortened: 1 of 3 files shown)


Guite:
Do I understand correctly that this is not going to be merged into master?

craigh:
> Do I understand correctly that this is not going to be merged into master?

technically, it will be merged to master but the legacy class is deleted there and the new class is already exactly the same (that is why the removals here - to align with master)

so in the end, no, master will not change, but yes, it will be merged ;-)

Guite:
Another question: why is the hook collector removed from the new class? Does this cause the new dispatcher failing? Did you test non-persisted  hooks still working using FooHookModule?

craigh:
> Another question: why is the hook collector removed from the new class? Does this cause the new dispatcher failing? Did you test non-persisted hooks still working using FooHookModule?

because the new class is not used at all in 1.5

in 2.0 there is no use for non-persisted hooks

PR #3752, opened 2017-08-06
Search - Correcting Conditional Statement

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT

## Description

Correcting the conditional statement to parse active modules to search.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/CHANGELOG-1.5.md
+++ b/CHANGELOG-1.5.md
@@ -10,7 +10,7 @@ CHANGELOG - ZIKULA 1.5.x
     - ?
 
  - Fixes:
-    - ?
+    - Correcting conditional statement for searchable modules.
 
  - Features:
     - ?
--- a/src/system/SearchModule/Api/SearchApi.php
+++ b/src/system/SearchModule/Api/SearchApi.php
@@ -85,7 +85,7 @@ class SearchApi implements SearchApiInterface
             $words = ($searchType == 'EXACT') ? [trim($q)] : preg_split('/ /', $q, -1, PREG_SPLIT_NO_EMPTY);
             $searchableModules = $this->searchableModuleCollector->getAll();
             foreach ($searchableModules as $moduleName => $searchableInstance) {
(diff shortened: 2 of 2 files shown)


craigh:
please add a changelog entry.

craigh:
how did you notice this problem? I'd like to understand better since there is no associated ticket. What caused the issue?

shefik:
I came across the bug while executing a search with my own custom modules. I noticed that all module's search helper code were being run. Troubleshooting the code revealed the problem with the conditional statement being incorrect.

So, to duplicate the issue, you will need to install a module that has search capabilities. You won't be able to encounter the bug with just the Zikula core alone, since it doesn't have search modules to test with.

shefik:
Would this really need a changelog entry? It's not a new feature. It's just a regular bug fix for something that was already broken.

craigh:
> You won't be able to encounter the bug with just the Zikula core alone, since it doesn't have search modules to test with.

The core allows searching of users, so there is one module.

My concern then (because last I checked it worked with the core) is if something may be wrong elsewhere then (in Users for example). 

> Would this really need a changelog entry

yes please.

shefik:
Yes, the Users module is indeed searchable. But, I was able to more easily and quickly spot the issue by having multiple searchable modules installed, and I noticed that I kept seeing results from all the modules.

craigh:
OK. Can you please test whether Users still functions as expected with your fix? 🙇

shefik:
Yes, Users still works.

craigh:
> Yes, Users still works.

great! just update the changelog and we can merge this then.

shefik:
Changelog has been updated.

craigh:
thank you @shefik

PR #3758, opened 2017-08-07
Sortablecolumns

### LINKED ISSUE #3757: [SortableColumns] still using legacy.css classes
| Q                 | A
| ----------------- | ---
| Zikula Version    | 2.0.0
| PHP Version       | 5.5.26

### Expected behavior
sortable columns display headers correctly

### Actual behavior
no styling at all

### Steps to reproduce
load any list view implementing SortableColumns

The problem is that the Column class assumes availability of legacy css classes

    class Column
    {
        // ...
        const CSS_CLASS_UNSORTED = 'z-order-unsorted';
        const CSS_CLASS_ASCENDING = 'z-order-asc';
        const CSS_CLASS_DESCENDING = 'z-order-desc';

and the file `src/style/legacy.css` which defines these classes is no longer included with Core-2.0

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | fixes #3757
| Refs tickets      | -
| License           | MIT

## Description
this is probably nearly the simplest solution

### DIFF
--- a/src/lib/Zikula/Bundle/CoreBundle/Resources/public/css/core.css
+++ b/src/lib/Zikula/Bundle/CoreBundle/Resources/public/css/core.css
@@ -260,3 +260,20 @@ input:invalid, select:invalid, textarea:invalid, .form-error {
 .zikula-bootstrap-tab-container .nav-tabs li {
     margin: 0 0 -1px 0;
 }
+
+.sorted-desc:after,
+.sorted-asc:after,
+.unsorted:after {
+    font-family: 'FontAwesome';
+    font-style: normal;
+    font-weight: normal;
+}
+.sorted-asc:after {
+    content: "\A0\f0de";
+}
+.sorted-desc:after {
+    content: "\A0\f0dd";
+}
+.unsorted:after {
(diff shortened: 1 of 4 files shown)


src/lib/Zikula/Bundle/CoreBundle/Resources/public/css/core.css
```diff
@@ -260,3 +260,20 @@ input:invalid, select:invalid, textarea:invalid, .form-error {
...
+
+.sorted-desc:after,
+.sorted-asc:after,
+.unsorted:after {
+    font-family: 'FontAwesome';
+    font-style: normal;
+    font-weight: normal;
+}
+.sorted-asc:after {
+    content: "\A0\f0de";
```
Guite: Sure the `\A0` prefix is correct?

craigh: no, but I think it is working. I was trying to get a space. open to better suggestions.

Issue #3745, opened 2017-08-03
Hooks issues


Kaik:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.4, 1.5, 2.0
| PHP Version       | 7.x

## Mixed notify hooks displaying methods
This one does not work. https://github.com/zikula/core/blob/1.5/src/system/ZAuthModule/Resources/views/Authentication/UnameLogin.html.twig#L26

### Expected behavior
Modify user display hooks informed about object id and object url
### Actual behavior
Modify user display hooks not informed about object id and object url
### Steps to reproduce
... long story short
https://github.com/zikula/core/blob/master/src/system/UsersModule/Resources/views/UserAdministration/modify.html.twig#L50

Kaik:
I will try all core modules that provide hook capabilities if hooks calls are working during Dizkus hooks refactor. So please do not close this ticket yet.

craigh:
status on this? I don't understand the ticket at all.

Guite:
> This one does not work.
Assigning `notifyDisplayHooks` and iterating over it is not possible. See also https://github.com/Guite/MostGenerator/commit/eb08e3b212394711284888e23ec726b194ca72a9

> not informed about object id and object url
urlObject parameter is missing on some calls

craigh:
closed by #3790

PR #3790, opened 2017-08-26
hook fixes refs #3745

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3745 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

ping @Guite @Kaik is this sufficient to fix the problem?

### DIFF
--- a/src/system/BlocksModule/HookSubscriber/UiHooksSubscriber.php
+++ b/src/system/BlocksModule/HookSubscriber/UiHooksSubscriber.php
@@ -17,6 +17,8 @@ use Zikula\Common\Translator\TranslatorInterface;
 
 class UiHooksSubscriber implements HookSubscriberInterface
 {
+    const HTMLBLOCK_EDIT_FORM = 'blocks.ui_hooks.htmlblock.content.form_edit';
+
     /**
      * @var TranslatorInterface
      */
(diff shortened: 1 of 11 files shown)


Guite:
@Kaik meant to review "all core modules". This PR only addresses Users. Either the other modules are missing or there were no errors there :wink:

craigh:
> Either the other modules are missing or there were no errors there

there are no other calls to `notifyDisplayHooks()` in any other templates.

Guite:
How about introducing them for (at least) Blocks and Groups?

Kaik:
Hi, sorry guys I'm really busy... I dive into Dizkus hooks so much I forgot to do this... Hooks are not really used in core... and looks like only UsersModule and BlockModule uses hooks - both do not have proper display_view call for object. UsersModule because there is no - user view page and BlockModule - hmm should be in block template.

craigh:
> How about introducing them for (at least) Blocks and Groups?

> hmm should be in block template

the html block has this hook call which I missed before but needs no adjustment:

https://github.com/zikula/core/blob/6ecc9936856be64eb603efd160c46e23c82494a5/src/system/BlocksModule/Resources/views/Block/html_modify.html.twig#L2

> only UsersModule and BlockModule uses hooks - both do not have proper display_view call for object

I would say that they are implemented properly. Please explain what you mean. 

> UsersModule because there is no - user view page

and there shouldn't be. That is handled by Profile.

@Guite if you desire further additions, please create a FR ticket and schedule it for 2.1.0. #3745 is only a bug fix.

PR #3813, opened 2017-09-21
use UrlInterface instead of string for search results

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

### DIFF
--- a/CHANGELOG-1.5.md
+++ b/CHANGELOG-1.5.md
@@ -8,6 +8,7 @@ CHANGELOG - ZIKULA 1.5.x
 
  - Fixes:
     - Allow hooks to be managed in display (#3793).
+    - Fixed broken search result url handling for new-style modules.
 
  - Vendor updates:
     - ?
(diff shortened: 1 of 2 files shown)


Portugao:
👍

Guite:
This PR ensures that `UrlInterface` instances are passed to the search results template instead of converting them to a string. Because the `generateUrl` Twig filter expects an `UrlInterface` and a string causes an error: `An exception has been thrown during the rendering of a template ("Catchable Fatal Error: Argument 1 passed to Zikula\SearchModule\Twig\TwigExtension::generateUrl() must implement interface Zikula\Core\UrlInterface, string given...").`

shefik:
@Guite Although that bug is fixed, I am now experiencing this error, when I apply the code from this pull request:

```An exception has been thrown during the rendering of a template ("Catchable Fatal Error: Object of class Zikula\Core\RouteUrl could not be converted to string").```

PR #3815, opened 2017-09-21
Get correct category by path, using the parent ID.

shefik:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT

## Description

When attempting to get a category by path, it is possible that multiple categories exist with the same name (although their parents differ). Therefore, in order to get the correct category by path, then the query must include the parent ID. Without also querying with the parent ID, then you may result in getting the incorrect category by just the non-unique name.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/src/system/CategoriesModule/Api/CategoryApi.php
+++ b/src/system/CategoriesModule/Api/CategoryApi.php
@@ -229,6 +229,23 @@ class CategoryApi
         $values = [];
         foreach ($apath as $path) {
             $parts = explode('/', $path);
+
+            if ($pathField == 'path') {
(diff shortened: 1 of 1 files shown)


Guite:
Please fix StyleCI issue and add a changelog entry.

Guite:
Added in https://github.com/zikula/core/commit/e661d1c492756237321ea7801928f2c988c62cec

PR #3837, opened 2017-10-30
No attr error

Kaik:
In case of an attribute with a $name does not exist return empty string.

| Q                 | A
| ----------------- | ---
| Bug fix?          | [yes]
| New feature?      | [no]
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | [no]

## Description
Fix error that will occur in case of not existing attr.

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Changelog

### DIFF
--- a/CHANGELOG-1.5.md
+++ b/CHANGELOG-1.5.md
@@ -7,6 +7,7 @@ CHANGELOG - ZIKULA 1.5.x
     - ?
 
  - Fixes:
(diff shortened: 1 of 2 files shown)


Guite:
Please add a changelog entry.

PR #3882, opened 2018-01-01
Translatable fallback

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

### DIFF
--- a/CHANGELOG-1.5.md
+++ b/CHANGELOG-1.5.md
@@ -10,6 +10,7 @@ CHANGELOG - ZIKULA 1.5.x
  - Fixes:
     - Fixed wrong request service call in GroupsModule menu (#3874).
     - Fixed fetching module url from metadata when untranslated (#3876).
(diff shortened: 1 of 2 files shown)


craigh:
I do not understand. what bug is this fixing?

Guite:
@craigh 

scenario:

- you have a page with for example English and German locales
- you want to create a new item with translatable fields (e.g. title and description)
- only the current language fields are mandatory (because not all speakers of language A are also able to speak language B)
- let's say you create a new item with current locale set to English but do not enter anything into the German translation fields

Without this PR the item's title and description are empty when the item is displayed in German.
With this PR there is a fallback to the main locale (English) version if no translation is saved.

Understandable?

Issue #3982, opened 2019-11-13
mmenu component name changed


Guite:
See https://github.com/FrDH/mmenu-js/blob/master/composer.json

Issue #3962, opened 2019-03-05
Use Flex for custom composer tasks


Guite:
https://github.com/sensiolabs/SensioDistributionBundle is deprecated in favour of Symfony Flex

Currently composer correctly installs vendors, but scripts fail afterwards.
```
Compiling component files
PHP Fatal error:  Class 'Sensio\Bundle\DistributionBundle\Composer\ScriptHandler' not found in /.../src/lib/Zikula/Composer/CopyRequirementsFile.php on line 18
```

This includes #3908

craigh:
the error is about copy the requirements checking file from the old symfony installer. (there is no relation to the component installer which is front end assets only).

You could probably use this instead

https://github.com/symfony/requirements-checker

Guite:
You are probably right about this specific script. The issue is bigger though. With Sf4 Flex is used for doing several tasks. So basically everything shown in https://github.com/zikula/core/blob/master/composer.json#L29-L35 is deprecated.

craigh:
well, I certainly haven't kept up with anything. I don't really understand flex at all. But it seems to me that composer will still run scripts. So I don't think you have to replace those lines but only find a replacement for the scripts from SensioDistributionBundle (because it is deprecated). But it is entirely possibly I don't know what I'm talking about.

craigh:
so - just to keep kicking this... I think what I am suggesting is that Zikula shouldn't be using flex (at least in Core 3). You can still use Symfony 4 without it. See Xabbuh's comment here:

https://stackoverflow.com/questions/48410209/where-are-docs-for-symfony4-without-flex

Kaik:
I think it is time to switch to encore and webpack and manage all Zikula assets via webpack. It will handle all of those scripts and more apart from the clear cache one.

It is really easy to install it can be installed in various ways and it can handle all assets, files, css, js, images, sass... We can do it now with legacy jquery and everything else in legacy mode.

craigh:
flex is about backend scripts (php) not front-end assets.

Kaik:
Yes, I know I meant those scripts in composer.json are about assets processing (bootstrap, js etc..), so everything apart from cache clear can be done by encore/webpack.

craigh:
I am looking into solutions for this ticket.

craigh:
@Guite closing this issue as fixed. The issues with SensioDistributionBundle are resolved. Further use of Flex is outside the scope of this ticket.

Issue #3987, opened 2019-11-16
PSR-12 support


Guite:
Since PSR-2 is deprecated consider switching to PSR-12

See also:

- https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-12-extended-coding-style-guide-meta.md#5-changelog-from-psr-2
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4502
- https://github.com/karriereat/php-code-quality/pull/9

Guite:
Reminder: put `declare( strict_types=1 );` below the file comment block.

Guite:
https://docs.styleci.io/presets#psr12

Issue #3644, opened 2017-07-14
combine modules and themes as one extension in DB


craigh:
Modules and themes should be considered one 'extension' in the DB. This means sharing most type of entity properties including capabilities, etc. 

event names should be updated here and replaced `EXTENSIONS_INSTALL` etc. and act upon themes as well.
 - `\Zikula\ThemeModule\EventListener\ExtensionInstallationListener::getSubscribedEvents`

craigh:
fixed by #4103

PR #3991, opened 2019-11-25
Installer fix

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | maybe?
| Deprecations?     | no
| Refs tickets      | #3962
| License           | MIT
| Changelog updated | not yet

## Description
trying to get project to build - e.g. composer will full work and run all scripts
also make sure that console runs properly

## Todos
- [x] mmenu is currently removed and must be re-added (probably refs #3982)
- [x] Liip/ImagineBundle is currently removed (error: Root image path not resolvable)
- [x] session.storage override in console is disabled refs 932f563fdaa9ae7b54289cc858f82a1bf1c7d267
- [x] probably other stuff.


craigh:
@Guite - your feedback is welcome. I will continue to work on this branch as I find time. You're assistance with the ImagineBundle would be very helpful. 

Of course, there are multiple things wrong here, this is only a work in progress.

src/lib/Zikula/Bundle/CoreInstallerBundle/Util/ZikulaRequirements.php
```diff
@@ -13,9 +13,9 @@
 
 namespace Zikula\Bundle\CoreInstallerBundle\Util;
 
-use Requirement;
+use Symfony\Requirements\Requirement;
 use Symfony\Component\Intl\Exception\MethodArgumentValueNotImplementedException;
```
Guite: Please sort imports alphabetically.

src/lib/Zikula/Bundle/CoreInstallerBundle/composer.json
```diff
@@ -14,9 +14,17 @@
...
         "symfony/framework-bundle": "4.*",
+        "symfony/console": "^4.3",
+        "symfony/requirements-checker": "dev-patch-1 as 1.*",
         "matthiasnoback/symfony-console-form": "3.*",
         "zikula/wizard": "3.*"
     },
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/craigh/requirements-checker"
```
Guite: What is the reason for using your fork? You did no changes there so far.

Guite: Found it: https://github.com/craigh/requirements-checker/commit/7149dcfdac06eec8c6f7efc0b1e2ecb4e0d25261

src/lib/Zikula/Core/Controller/AbstractController.php
```diff
@@ -108,7 +108,7 @@ protected function decorateTranslator(array $parameters): array
      * Returns a NotFoundHttpException; this will result in a 404 response code.
      * Usage example: throw $this->createNotFoundException();
      */
-    public function createNotFoundException(string $message = 'Not Found', ?Throwable $previous = null): NotFoundHttpException
+    public function createNotFoundException(string $message = 'Access Denied.', \Throwable $previous = null): NotFoundHttpException
```
Guite: Please change `Access Denied.` back to `Not Found` like shown at https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/FrameworkBundle/Controller/ControllerTrait.php#L291

src/system/ThemeModule/EventListener/DefaultPageAssetSetterListener.php
```diff
@@ -68,7 +68,7 @@ public function __construct(
         string $bootstrapJavascriptPath,
         string $bootstrapFontAwesomeStylesheetPath,
         string $fontAwesomePath,
-        string $bootstrapStylesheetPath
+        $bootstrapStylesheetPath // cannot cast to bool because set with expression language
```
Guite: Please replace `bool` by `string`.

Guite:
I like the approach: getting the composer instance from the event seems a nice to migrate our custom actions.

craigh:
@Guite can you help with this a bit? A lot of it is working now. When in `dev` mode, If you do `bin/console z:i:s` and complete it, then `bin/console z:i:f` you will see that there is a memory error when it is trying to install the Theme module. I can't figure it out.

Also, perhaps you can figure out the problem with ImagineBundle. I could not get the configuration correct (error was `Root image path not resolvable`). Since I don't use Imagine, I don't even know where to begin. I tried something like `data_root: '%kernel.root_dir%/../web/'` with no success.

Guite:
I've added a few fixes.

> memory error when it is trying to install the Theme module
disable `dev` mode, the memory problem is probably caused by the profiler; or set `memory_limit=-1` for testing

one guess is that the theme table contains fields named `type` and `system` which could make problems being a reserved word...

```
Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'system SMALLINT NOT NULL, state SMALLINT NOT NULL, xhtml SMALLINT NOT NULL, PRIM' at line 1' while executing DDL: CREATE TABLE themes (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(64) NOT NULL, type SMALLINT NOT NULL, displayname VARCHAR(64) NOT NULL, description VARCHAR(255) NOT NULL, version VARCHAR(10) NOT NULL, contact VARCHAR(255) NOT NULL, admin SMALLINT NOT NULL, user SMALLINT NOT NULL, system SMALLINT NOT NULL, state SMALLINT NOT NULL, xhtml SMALLINT NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET UTF8 COLLATE `UTF8_unicode_ci` ENGINE = InnoDB
```

Also the extensions table contains a field named `type`.

Maybe worth renaming two fields to `extensionType`, `themeType` and remove `system` entirely? slightly refs #3644


Also at the very end of the CLI install finish process I get:

```
Finalize parameters
 Finalizing parameters

In getThemeEntityRepositoryService.php line 3:
                                                              
  Maximum function nesting level of '256' reached, aborting!
```

Guite:
> I've added a few fixes.

Sorry, forgot to push. Will do later.

craigh:
✅ I am using prod mode apparently (thought it was dev until I checked)
❌ Still getting same 'out of memory' errors despite most recent commit that fixes reserved word issue.

I'm guessing the `Maximum function nesting level of '256' reached, aborting!` is the real problem. Somehow we must have a memory leak or much deeper nesting of objects than I thought. Do you have other ideas?

craigh:
@Guite can you please take a look at this? I've been staring at it for hours and not coming up with a solution.

Basically, the problem is that Twig is not registering any custom extensions. The services are present and properly autowired/configured and include `twig.extension` tag as needed. for example `bin/console debug:container ModVarExtension` produces expected result. but `bin/console debug:twig` will not show the functions registered in said extension. And `bin/console lint:twig` shows that all the custom functions/filters/etc are not recognized.

I'm hoping I missed something obvious. thanks!

Guite:
> Basically, the problem is that Twig is not registering any custom extensions.

Do you refer to the decorated events? I experienced similar problems before when I worked on #2221 in last spring - you might read through all links referenced in that ticket. But I also think this is outside of the scope of this PR.

craigh:
> Do you refer to the decorated events?

no. I simply mean that Twig cannot find custom extensions. Look at the errors in the travis build, you will see what I mean.

> But I also think this is outside of the scope of this PR.

The scope of the PR is to get the project to build without error. It is also to get both the CLI and web installer to fully function producing at least a first page error free.

Guite:
Okay. I thought you wanted to include #2221 entirely.

> I simply mean that Twig cannot find custom extensions.

Auto configuration of Twig-related service tags is done in the TwigBundle:
https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php#L144

Is this bundle not available anymore?

Then we could do this in one of our bundles.
Examples:
- https://github.com/zikula/core/blob/master/src/lib/Zikula/Bundle/CoreBundle/DependencyInjection/CoreExtension.php#L42
- https://github.com/zikula/core/blob/master/src/lib/Zikula/Bundle/HookBundle/DependencyInjection/ZikulaHookExtension.php#L33
- https://github.com/zikula/core/blob/master/src/system/UsersModule/DependencyInjection/ZikulaUsersExtension.php#L31

I would recommend doing this in the Theme module then.

craigh:
we shouldn't need to do this. The TwigBundle is present and some services from it are in use. you can use `bin/console debug:container` to see what is there and registered.

craigh:
yes #2221 will need to be fully included here I think

src/system/ThemeModule/Resources/config/services.yml
```diff
@@ -84,10 +95,11 @@ services:
           $combine: '%zikula_asset_manager.combine%'
 
     Zikula\ThemeModule\Bridge\Twig\EventEnabledTwigEngine:
-        decorates: templating.engine.twig
+        decorates: twig
         arguments:
-          $parser: '@templating.name_parser'
-          $locator: '@templating.locator'
+          $loader: '@twig.loader'
```
craigh: @Guite given your previous comments, perhpahs this is the problem. Should `$loader` be something else?

Guite: No, I think the missing extensions are not related to any type of template loader.

src/system/RoutesModule/AppSettings.php
```diff
@@ -1,17 +1,17 @@
 <?php
 
-declare(strict_types=1);
```
craigh: in the rest of the core, the declaration is at the top (like this) not after the comment, like below. Why change it?

Guite: Demanded by phpcs / PHP_CodeSniffer. Maybe to conform to PSR-12 standard.

Also it we are actually documenting the file, not the `namespace` keyword. Thus, the comment must come first. See also https://github.com/squizlabs/PHP_CodeSniffer/issues/1800#issuecomment-352634571

You don't need to mind about it. I will change rest of the core later. Refs #3987

craigh:
Screen Shot 2019-12-14 at 9 19 45 AM

craigh:
regarding 838fa1c4407fb0fcaaf3b033dcc425608ab97784

for reference:
Screen Shot 2019-12-14 at 8 48 38 AM

paths can be tricky. especially because different production environments can be in myriad sub-directories.

Issue #4032, opened 2019-12-31
[Theme] [RFC] Strategy about module/theme and asset management


Guite:
## Background and context
I looked into #3908 and started reading. The problem with this issue is that when you start replacing the component-installer you immediately end at the question how we treat assets and libraries in general.

1. General libraries (in the core)
2. Module / theme / bundle specific assets

For the 1st part the component-installer uses `"component-dir": "public"` in our `composer.json` file and each component (jquery, jquery-ui, etc.) specifies which assets to process ([example](https://github.com/components/bootstrap/blob/4/composer.json#L30)).

## What is missing
Beside the component-installer being deprecated we still have no real asset management nor support for yarn and all this stuff (see #3946). We are also not using Webpack Encore yet (see #3601). Which is of course not easy in a system which consists of several different modules etc. Also we want/need that connected to composer (see #3273, #3438, #3804, Symfony Flex). There are also thoughts about unifying modules and themes (see #3644) and treating them as vendors in future.

Time to look what other's do.

## Alternatives

The component-installer readme shows three alternatives:

1. https://github.com/oomphinc/composer-installers-extender
2. https://github.com/hiqdev/asset-packagist
3. https://github.com/fxpio/composer-asset-plugin

The third one is the most popular one. But it's new version is incompatible with older ones and became a new project:

4. **https://github.com/fxpio/foxy**

## How about other projects?

Other PHP projects have similar requirements. Here is some related reading material:

- https://github.com/drupal-composer/drupal-project/issues/278
- https://github.com/yiisoft/assets/issues/7
- https://github.com/fxpio/composer-asset-plugin/pull/330

## What next

For all these changes we need to agree about the direction we are heading to. Otherwise it is difficult to prioritise things and tackle them one by one.

Kaik:
It might be irrelevant or off topic but...
So...
Copying files from one place to another during installation is not the problem I guess.

Both composer and yarn/npm are dev tools. Both of them are designed and used to compose/build sources into artifacts, which are then tested and deployed.

When build is deployed it should not change.

While we (and other cms) want to add additional packages during production which means, we want to recompile stuff in production.

At the moment our 3rd party modules are bypasing composer when you install them via web, and this works because PHP is mostly interpreted "on the fly". If someone needs 3rd party php lib in his zikula module there is no way to install it via composer, you have to supply it with the module(which is bad).

[...]

Issue #1832, opened 2014-06-29
Use original packages in composer


phaidon:
For some composer packages we are using own bundles. E.g. zikula/jquery-bundle

But I think there is no reason for that. The path (src/web) works without the original bundles and the bundles don't do anything additional (they are empty packages with a dependency to original packages).

So I suggest to use just the original packages. Or do I miss something?

@drak @craigh @cmfcmf 

Here the affected packages: zikula/jquery, zikula/jquery-ui, zikula/jquery-minicolors-bundle, zikula/bootstrap-bundle, zikula/fontawesome-bundle

craigh:
After doing about an hour of research, it appears that all the proposals at composer were eventually denied and composer remains a php-only solution. Additionally, it also appears that bower is [losing its dominance](https://gofore.com/stop-using-bower/) in the js world and npm is gaining momentum.

craigh:
but this is maybe the descendant of those discussions in composer: https://github.com/francoispluchino/composer-asset-plugin

cmfcmf:
Maybe we should reconsider using NPM / Bower.

Guite:
Probably related and interesting here: http://www.slideshare.net/nachomartin/webpacksf

Guite:
Added blocker tag because I also think that our custom bundle packages are too cumbersome.

Guite:
> javiereguiluz [9:27 AM]
> for Symfony apps I recommend you “webpack" and “yarn"

See also: https://www.youtube.com/watch?v=iINLewCGnuY

craigh:
so - the reason we do this is because the original vendors are not symfony bundles. This means that there is no automatic management of web assets (images, css and js files).

Symfony will automatically copy **bundle** contents of `MyFooBundle/Resources/*` to `/web/bundles/myfoo/*` then these assets can be easily accessed in a template with `{{ asset('bundles/myfoo/images/image.jpg') }}`. This is important because ideally, we would like to make only the `/web` directory publicly viewable - assets that remain in `/vendors` or `/modules` are by default *not* visible to a browser - causing the site to break.

[...]

PR #4033, opened 2019-12-31
Removed dummy bundles

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #1832 
| Refs tickets      | #4032 
| License           | MIT
| Changelog updated | yes


Guite:
@craigh I don't see any paths which need amendments. Can you explain which ones you are referring to?

craigh:
try manually clearing your `/web` folder and running `composer install`. See if your site works (e.g. all the assets load properly). If not, you should probably look at (at minimum) these paths:

https://github.com/zikula/core/blob/5771fadd5eae37d5cbca8286ee8ab8e45eff7ac2/src/system/ThemeModule/EventListener/DefaultPageAssetSetterListener.php#L126-L127

https://github.com/zikula/core/blob/46ea02f9675ea65b692410fd52a6380144ba2569/src/lib/Zikula/Composer/LessGenerator.php#L34

Guite:
It all works. Actually no paths really changed...

craigh:
really. that seems impossible. but if you say so 😄

Guite:
Well, the paths are still defined by `components/foo`. The only difference is that we now use these directly and not indirectly through our dummy bundles.

Issue #3530, opened 2017-04-06
Upgrade to Bootstrap 4.0


craigh:
Upgrade to Bootstrap 4.0

See PR at #4037

Guite:
BS4 beta arrived:
https://github.com/twbs/bootstrap/releases/tag/v4.0.0-beta

Symfony 3.4 will get a BS4 form theme:
https://github.com/symfony/symfony/pull/21751

Guite:
http://symfony.com/blog/new-in-symfony-3-4-bootstrap-4-form-theme

robbrandt:
I am working with a bootstrap 4 theme now and have found that 4 breaks many bs 3 plugins, and bs 4 plugins haven't been developed yet. It may be wise to proceed slowly.

craigh:
> It may be wise to proceed slowly.

that is why this ticket is assigned to the 3.0.0 milestone. which is (at best) several years away.

craigh:
Bootstrap 4 released 18 January 2018
https://blog.getbootstrap.com/2018/01/18/bootstrap-4/

Guite:
http://symfony.com/blog/new-in-symfony-4-1-improved-bootstrap-4-form-theme

Guite:
4.1.0 arrived today https://github.com/twbs/bootstrap/releases/tag/v4.1.0

craigh:
http://upgrade-bootstrap.bootply.com

https://designmodo.com/migrate-bootstrap-4/

Guite:
Fixed in #4037

PR #4037, opened 2020-01-02
Upgrade to Bootstrap 4

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | yes
| Deprecations?     | no
| Fixed tickets     | #3530
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description
As part of this PR Bootstrap is upgraded from version 3 to version 4. Also we are simplifying it's integration into the core.

Particularly we completely remove the less/sass meta layer for these main reasons:

1. Much more complexity than needed: basically we use Bootstrap, but we don't extend or amend it significantly.
   - We want to stay close to the standard.
   - The scss files are really many and we just don't want to fiddle around with them.
2. More update pain: the component repo does not include any sass/scss sources atm (for similar reasons).
3. Bootstrap 4 does not contain glyph icons anymore (which we usually removed in favour of font awesome). Instead they [recommend using FA now](https://getbootstrap.com/docs/4.3/extend/icons/#preferred), too.

So from now on we just treat Bootstrap as what it is - a vendor component - and work with it instead on it.

If somebody wants to further customize things he can still do it. We provide the possibility to use a custom bootstrap css path for that. But Zikula does not have to provide the "source" experience of the whole BS package ootb.

This PR also removes the "zikula.stylesheet.bootstrap-font-awesome.path" because there is no need to have this in one file then anymore. Asset combination can do this instead if desired.

## Todos
- [x] Check that latest [BS release](https://github.com/twbs/bootstrap/releases) is available in [component repo](https://github.com/components/bootstrap/releases) (created PR: https://github.com/components/bootstrap/pull/59)
- [x] Completely remove the less/sass meta layer
- [x] Remove the "zikula.stylesheet.bootstrap-font-awesome.path"
- [x] We have some simple cosmetic overrides in our bootstrap_variables. Review them and either remove them or possibly put them into core.css for the moment.
- [x] Check if [modal fix for mmenu](https://github.com/zikula/core/blob/master/src/system/AdminModule/Resources/public/js/ZikulaAdminModule.Modal.Fix.js) is still needed.
- [x] Adapt `FormExtensionBundle` to Symfony's Bootstrap 4 Form Theme
- [x] Upgrade bundle and module templates
- [x] Remove Andreas08 and SeaBreeze themes
- [x] Upgrade Bootstrap theme
- [x] Extend Bootstrap theme to support not only cerulean multiple themes from https://bootswatch.com/ - at the end these are only [different css files](https://github.com/thomaspark/bootswatch/tree/v4.4.1/dist) and they [support composer](https://github.com/thomaspark/bootswatch/blob/master/composer.json)
- [x] Provide an additional Bootstrap theme variable (checkbox option) to allow users switching between Bootswatch styles using a dropdown. Unsure whether it's worth the effort (maybe something for later).


Guite:
Removal of old stuff:

Screenshot_20200105_162113


Style switcher:

Screenshot_20200105_162136

Guite:
"Sketchy":

Screenshot_20200105_174105

craigh:
I am getting this error:
> [Error] Failed to load resource: the server responded with a status of 404 (Not Found) (bootstrap.bundle.min.js.map, line 0) http://127.0.0.1/core.git/src/web/bootstrap/js/bootstrap.bundle.min.js.map

Guite:
> I am getting this error:
> 
> > [Error] Failed to load resource: the server responded with a status of 404 (Not Found) (bootstrap.bundle.min.js.map, line 0) http://127.0.0.1/core.git/src/web/bootstrap/js/bootstrap.bundle.min.js.map

Will be fixed here: https://github.com/components/bootstrap/pull/59/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

Issue #4052, opened 2020-01-09
Convert directory structure and configuration to Sf5 default


Guite:
The following tasks would be relevant for getting closer some best practices, like [standard directory layout](https://symfony.com/doc/current/best_practices.html#use-the-default-directory-structure) and using/supporting [environment variables](https://symfony.com/doc/current/best_practices.html#use-environment-variables-for-infrastructure-configuration).

- [x] 1. Rename/move config directory and files (done in #4079, see changes [here](https://github.com/zikula/core/pull/4079/files#diff-8ee894204f471eae7d606c828a1af52f))
- [x] 2. Rename/move other directories and files (done in #4079, see changes [here](https://github.com/zikula/core/pull/4079/files#diff-8ee894204f471eae7d606c828a1af52f))
- [x] 3. Further changes to directory layout (done in #4084, see changes [here](https://github.com/zikula/core/pull/4084/files#diff-8ee894204f471eae7d606c828a1af52f))
- [x] 4. Post processings (done in #4087)

craigh:
perhaps we could slack on Saturday and discuss?

Guite:
Yes. If you have two minutes later, I'd like to coordindate something else with you briefly though.

craigh:
about 3pm EST

craigh:
While [this page](https://symfony.com/doc/current/bundles.html) discourages bundles at all, like it or not, Zikula _will_ be a collection of bundles. I think we can eventually put CoreBundle, CoreInstallerBundle, FormExtensionBundle, HookBundle and WorkflowBundle into vendors as a composer requirement. Now that I [moved ](401c464) bootstrap.php, this leaves only RequirementChecker in `/lib`. For the moment, I propose this:

    src/
        Zikula/
            CoreBundle/
            CoreInstallerBundle/
            FormExtensionBundle/
            HookBundle/
            WorkflowBundle/
            RequirementChecker.php
            Kernel.php

this eliminates the `/lib` and also `/Bundles` sub directories in that path and aligns with Sy5 best-practices

In the distribution, the Bundles would move to `/vendor` leaving only RequirementChecker.php & Kernel.php

I think this can all be done immediately and continue with a working install.

Guite:
> While this page discourages bundles at all, like it or not

Initially I misunderstood this, too. When Symfony 4 arrived I asked about it in the Symfony Slack. The answer was that bundles are still fine for reusable components which are distributed.
[...]

PR #4084, opened 2020-01-26
updated directory layout

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | yes
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | #4052
| License           | MIT
| Changelog updated | yes

## Description
This PR changes the directory layout like shown in [this comment](https://github.com/zikula/core/issues/4052#issuecomment-578354992).
Therewith it covers step 3 of #4052 

It also updates the logic of how assets are resolved (looking first in `public/overrides/`). But this needs to be reviewed in the next iteration when `index.php` and document root are moved into `public` (step 4 of #4052).


craigh:
actually - `index.php` is supposed to be in `public`

Guite:
> actually - index.php is supposed to be in public

Postponed to next step.

PR #4088, opened 2020-01-28
Fixed paths for adding autoloaders

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description

This PR ensures that PSR4 namespaces are added with the correct path.

Background is that from the class loader's point of view classes are now not in `modules|themes/...` anymore, but `src/(modules|themes)/`.

### DIFF
--- a/src/Zikula/CoreBundle/Helper/PersistedBundleHelper.php
+++ b/src/Zikula/CoreBundle/Helper/PersistedBundleHelper.php
@@ -120,14 +120,15 @@ class PersistedBundleHelper
      */
     public function addAutoloaders(ZikulaHttpKernelInterface $kernel, array $autoload = []): void
     {
+        $srcDir = dirname(dirname(dirname(__DIR__))) . '/';
         if (isset($autoload['psr-0'])) {
             foreach ($autoload['psr-0'] as $prefix => $path) {
-                $kernel->getAutoloader()->add($prefix, $path);
+                $kernel->getAutoloader()->add($prefix, $srcDir . $path);
             }
         }
         if (isset($autoload['psr-4'])) {
             foreach ($autoload['psr-4'] as $prefix => $path) {
-                $kernel->getAutoloader()->addPsr4($prefix, $path);
+                $kernel->getAutoloader()->addPsr4($prefix, $srcDir . $path);
             }
         }
         if (isset($autoload['classmap'])) {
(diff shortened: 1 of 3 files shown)


src/Zikula/CoreBundle/Helper/PersistedBundleHelper.php
```diff
@@ -120,14 +120,15 @@ private function extensionIsActive(Connection $conn, string $class, string $type
      */
     public function addAutoloaders(ZikulaHttpKernelInterface $kernel, array $autoload = []): void
     {
+        $srcDir = dirname(dirname(dirname(__DIR__))) . '/';
```
craigh: this feel odd to me. what about using `$kernel->get ProjectDir()` again? Does this produce a full or relative path?

Guite: you are right; added usage of `getProjectDir()` in df8ab0cfe0f31a446451f3f23e29f317155bff24

src/Zikula/CoreBundle/Helper/PersistedBundleHelper.php
```diff
@@ -120,14 +120,15 @@ private function extensionIsActive(Connection $conn, string $class, string $type
      */
     public function addAutoloaders(ZikulaHttpKernelInterface $kernel, array $autoload = []): void
     {
+        $srcDir = $kernel->getProjectDir() . '/src/';
         if (isset($autoload['psr-0'])) {
             foreach ($autoload['psr-0'] as $prefix => $path) {
-                $kernel->getAutoloader()->add($prefix, $path);
+                $kernel->getAutoloader()->add($prefix, $srcDir . $path);
```
craigh: @Guite something else to consider - with this change, the autoloader paths are all full paths. In the past they were relative. Maybe this doesn't matter. What does composer generate for the other vendors, full or relative? And if they are relative, couldn't you just put src/ in here?

Guite: I looked into Composer's `autoload_psr4.php` and they are absolute there, too.

Issue #3955, opened 2019-02-04
multilingual start controller


rallek:
| Q                 | A
| ----------------- | ---
| Zikula Version    | 2.0.12
| PHP Version       | 7.1

### Expected behavior
In the start setting the start controller should be multilingual as well

### Actual behavior
![grafik](https://user-images.githubusercontent.com/935250/52237579-376e6f80-28ca-11e9-9d49-7301326ced9e.png)
only meta keywords are multilingual

### Steps to reproduce
install ContentModule and try to setup a page for German and a translation for English

rallek:
ups, If I use the same slug for german and english it seems to work propperly.
But nevertheless it would be a nice feature to be able to have a different start module setting for each language

Guite:
Of course this issue is a valid feature request.

rallek:
the value for this page is in both cases ``seiten/startseite`` and it is always showing the English page.

Guite:
@rallek please add the patch from https://github.com/zikula/core/commit/67b78dcbe7673f3c54c0d11906ed600d413f5f40 and try again...

Guite:
seems I added this during last camp btw ;-)

rallek:
same behavior. 

What should have changed? equal slug should language sensitive?

I deleted the folder ``prod`` inside cache after uploading the file.

rallek:
I am using zk 2.0.12. There this fix is already included. So I am able to set content as a startmodule.

rallek:
rechecked it right now. Might be it is working. Need further tests! Please keep this ticket open for some days until I can confirm.

rallek:
It is working fine now for ContentModule. But nevertheless it would be great to have the setting also multilingual.

craigh:
@Guite I don't understand this ticket. If you want to implement, please go for it.

Guite:
I am going to add the following changes:

- add support for GET, POST and Request parameters (like in Content's controller type, see [here](https://github.com/zikula-modules/Content/blob/master/extensions/Zikula/ContentModule/ContentType/Form/Type/ControllerType.php#L30-L57))
- add the possibility to enter this for each installed/available language with first fallback to default language and second fallback to nothing (static home page)

Guite:
Probably create a dedicated form type for centralising/grouping these four fields (controller, query, request, attributes) to simplify repeated usage.

PR #4122, opened 2020-02-08
extended start page settings for more options and different vars per language

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #3955 
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description

This PR includes the following changes:

- add support for GET and Request parameters for the start page controller
- centralise all related fields into a dedicated form type in `FormExtensionBundle` (controller, query, request, attributes) to simplify repeated usage
- a dropdown allows to choose a route/controller combination
- add the possibility to enter this for each installed/available language
- makes the `_route` argument properly available to the sub request

## Screenshot

Screenshot_20200210_185055

## Todo post-merge

- adaption/reuse in Content (also use shorthand syntax [here](https://github.com/zikula-modules/Content/blob/master/extensions/Zikula/ContentModule/ContentType/ControllerType.php#L127) and [here](https://github.com/zikula-modules/Content/blob/master/extensions/Zikula/ContentModule/ContentType/ControllerType.php#L165); also remove [this](https://github.com/zikula-modules/Content/blob/master/extensions/Zikula/ContentModule/ContentType/ControllerType.php#L137-L145))


src/Zikula/CoreBundle/Controller/MainController.php
```diff
@@ -46,22 +56,57 @@ public function __construct(ZikulaHttpKernelInterface $kernel, VariableApiInterf
...
+
+        $isValidStartController = true;
+        $startController = $startPageInfo['controller'];
+        if (!isset($startPageInfo['controller']) || !is_string($startPageInfo['controller'])) {
+            $isValidStartController = false;
+        } elseif (false === mb_strpos($startController, '\\') || false === mb_strpos($startController, '::')) {
+            $isValidStartController = false;
+        } else {
+            [$vendor, $bundleName] = explode('\\', $startController);
+            $bundleName = $vendor . $bundleName;
```
craigh: please use `$extensionName` instead of `$bundleName`

src/Zikula/CoreInstallerBundle/Helper/ParameterHelper.php
```diff
@@ -148,12 +148,27 @@ public function finalizeParameters(bool $configureRequestContext = true): bool
...
         if (isset($params['upgrading'])) {
             $params['zikula_asset_manager.combine'] = false;
-            $startController = $this->variableApi->getSystemVar('startController');
-            [$moduleName] = explode(':', $startController);
-            if (!$this->kernel->isBundle($moduleName)) {
-                // set the 'start' page information to empty to avoid missing module errors.
-                $this->variableApi->set(VariableApi::CONFIG, 'startController', '');
-                $this->variableApi->set(VariableApi::CONFIG, 'startargs', '');
+
+            // unset start page information if needed to avoid missing module errors
```
craigh: "if needed"... not sure this is a good idea. easier to simply unset it all and allow the site admin to reset manually.

Guite: sounds too much to do this on *every* update... maybe we could detect if it is a major upgrade?

src/Zikula/FormExtensionBundle/Form/Type/ControllerType.php
```diff
@@ -0,0 +1,106 @@
+<?php
```
craigh: IMO this should be back in settings module. it is very specific to that usage and not a general purpose "ControllerType" validator. Someone may come along later and wonder why it doesn't work for that purpose.

craigh: ah - it is not a validator - still the point remains valid. this is a very specific form type, not a general purpose. please move to SettingsModule

Guite: This isn't a validator, but a form type. At least Content will use it, too. It may useful whereever a sub request should become configurable.

src/Zikula/CoreInstallerBundle/Helper/ParameterHelper.php
craigh: disagree. It isn't _that_ much to do and is far safer.

Guite: changed like you wish

Issue #4115, opened 2020-02-06
make more public services private?


craigh:
https://github.com/zikula/core/blob/0e09cf99c2947552b5382a5e5ec6cb11e9eac763/src/system/ExtensionsModule/Resources/config/services.yaml#L22

is there any way to test some of these again? I think many of them no longer need to be public anyway due to refactoring since the original conversion.

Guite:
search for `get(TheType::class)` in whole `src/` directory?

craigh:
how did you figure it out in the first place when you refactored all that?

Guite:
Exceptions where thrown like "could not access xy service using container because it is private".

Guite:
A service can be made private again if it is only injected, but not accessed via container anymore.

PR #4125, opened 2020-02-10
Extension installer refactor

craigh:
remove more ContainerAware usages with respect to ExtensionInstaller classes. This makes other services not need to be public.

refs #4115

### DIFF
--- a/src/system/RoutesModule/Resources/config/services.yaml
+++ b/src/system/RoutesModule/Resources/config/services.yaml
@@ -19,6 +19,10 @@ services:
         resource: '../../Helper/*'
         lazy: true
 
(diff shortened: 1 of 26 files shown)


src/system/RoutesModule/Resources/config/services.yaml
```diff
@@ -19,10 +19,6 @@ services:
         resource: '../../Helper/*'
         lazy: true
 
-    # public because EntityLifecycleListener accesses this using container
-    Zikula\RoutesModule\Entity\Factory\EntityFactory:
-        public: true
-
```
Guite: Please revert the Routes change.

Guite:
Except the Routes change: :+1:  :+1:  :+1:

PR #4139, opened 2020-02-18
Module docs/help system

### LINKED ISSUE #3739: Module docs
| Q                 | A
| ----------------- | ---
| Zikula Version    | 2.x.x
| PHP Version       | 5.x

Hi, I just realized that someone I think @Guite have done nice help function for admin panel module.
I remember thinking about that long time ago and this help function could do a lot of good things like load files from docs directory there is another similar help function in extensions module. 

It's feature to discuss we could have nice docs for users, admins and developers (dev mode) but this requires that we/developers would need to add to ``Resources/docs`` folders like ``user admin dev `` in each module and place there some doc files (or something like that) so help function could read those and display. Other options like different languages etc.. are also possible. This is a general idea to discuss.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | fixes #3739
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description

This PR implements the innovations proposed in [this comment](https://github.com/zikula/core/issues/3739#issuecomment-587462105).

## Todo Post-Merge

- [ ] Add simple doc to VA modules
- [ ] Add simple doc to generator


docs/AccessControl/Authentication/ZAuthConfiguration.md
```diff
@@ -0,0 +1,113 @@
...
+to control how the extension handles various aspects of its functions and features.
+
+## User credential settings
+
+These options control various aspects of how the user logs into the system and
+how his account credentials are managed.
+
+### Minimum length for user passwords
+
+Default: 5
```
craigh: default is 8. minimum is 5

Guite: thanks! fixed

PR #4143, opened 2020-02-22
Add PermissionCheck Annotation

craigh:
```php
/**
 * @Annotation
 * @Target({"METHOD"})
 * Controller action permissions
 *
 * This annotation is used in a Controller Action Method in one of two ways.
 * 1. like so: @PermissionCheck("admin")
 *     Possible alias values are:
 *       'admin'
 *       'delete'
 *       'add'
 *       'edit'
 *       'moderate'
 *       'comment'
 *       'read'
 *       'overview'
 *     In the above cases,
 *       the component will be like 'AcmeFooModule::'
 *       the instance will be '::'
 *       the level will be the corresponding ACCESS_* constant (e.g. ACCESS_ADMIN)
 *     Also allowed: @PermissionCheck("ACCESS_ADMIN")
 *
 * 2. You can also pass any valid permission schema (e.g. @PermissionCheck({"ZikulaCategoriesModule::category", "ID::5", "ACCESS_EDIT"})
 *     - note the use of curly brackets `{}` within for this type of value
 *     The listener will attempt to replace any variable with a route attribute value. For example if this is the annotation:
 *       @PermissionCheck({"ZikulaGroupsModule::", "$gid::", "ACCESS_EDIT"})
 *     Then, the listener will look for an 'gid' attribute in the request object and replace the variable name with its value
 *     when testing for permissions.
 *     You can also use `$_zkModule` as the Extension name if preferred, e.g. @PermissionCheck({"$_zkModule::", "$gid::", "ACCESS_EDIT"})
 *     You can also use access alias if preferred, e.g. @PermissionCheck({"$_zkModule::", "$gid::", "edit"})
 *
 * @see \Zikula\PermissionsModule\Listener\ControllerPermAnnotationReaderListener
 */
```

### DIFF
--- a/src/system/PermissionsModule/Listener/ControllerPermAnnotationReaderListener.php
+++ b/src/system/PermissionsModule/Listener/ControllerPermAnnotationReaderListener.php
@@ -139,7 +139,7 @@ class ControllerPermAnnotationReaderListener implements EventSubscriberInterface
         if (!$this->hasFlag($segment)) {
             return $segment;
         }
-        if (1 !== preg_match('/\$([^:\n]+)/' , $segment, $matches)) {
+        if (1 !== preg_match('/\$([^:\n]+)/', $segment, $matches)) {
             throw new AnnotationException('Invalid schema in @Annotation: @PermissionCheck(). Could not match route attributes');
         }
 
@@ -153,7 +153,7 @@ class ControllerPermAnnotationReaderListener implements EventSubscriberInterface
         return $segment;
     }
 
-    private function filterMatches(string $value) : bool
+    private function filterMatches(string $value): bool
     {
         return !$this->hasFlag($value);
     }
@@ -166,10 +166,10 @@ class ControllerPermAnnotationReaderListener implements EventSubscriberInterface
     private function getValidConstant(string $string): int
     {
         if ($this->isValidConstantValue($string)) {
-            if (FALSE !== array_key_exists($string, $this->accessMap)) {
+            if (false !== array_key_exists($string, $this->accessMap)) {
                 return constant($string);
             }
-            if (FALSE !== $key = array_search($string, $this->accessMap)) {
+            if (false !== $key = array_search($string, $this->accessMap)) {
                 return constant($key);
             }
         }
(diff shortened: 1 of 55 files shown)


src/system/PermissionsModule/Listener/ControllerPermAnnotationReaderListener.php
```diff
@@ -0,0 +1,162 @@
...
+use Symfony\Component\HttpKernel\Event\ControllerEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Zikula\PermissionsModule\Annotation\PermRequired as PermAnnotation;
+
+class ControllerPermAnnotationReaderListener implements EventSubscriberInterface
+{
+    /**
+     * flag of route attribute parameter
+     */
+    const ROUTE_ATTRIBUTE_FLAG = '$';
```
Guite: please add a visibility modifier here

src/system/PermissionsModule/Listener/ControllerPermAnnotationReaderListener.php
```diff
@@ -0,0 +1,162 @@
...
+    {
+        if (3 !== count($schema)) {
+            return false;
+        }
+        if (array_sum(array_map('is_string', $schema)) !== count($schema)) {
+            return false;
+        }
+        if (((false !== mb_strpos($schema[0], ':')) && (2 !== mb_substr_count($schema[0], ':')))
+            || ((false !== mb_strpos($schema[1], ':')) && (2 !== mb_substr_count($schema[1], ':')))) {
+                return false;
```
Guite: You can use less paranthesis here:
```php
if (
    (false !== mb_strpos($schema[0], ':') && 2 !== mb_substr_count($schema[0], ':'))
    || (false !== mb_strpos($schema[1], ':') && 2 !== mb_substr_count($schema[1], ':'))
) {
```

Guite:
Are you going to add docs for that or should I include it into my upcoming general docs about permission system?

Guite:
Are you going to change more system modules to use it or is that going to be done in a separate step?

Guite:
Also please add a changelog entry.

PR #4144, opened 2020-02-22
composer cleanup; moved icon out of capabilities, fixed #4116

### LINKED ISSUE #4116: move icon out of capabilities
Now that extensions have a dedicated icon field in their database table, how about reflecting this in `composer.json`? This would also make the `BundleSyncHelper` a bit less complex again.

Current:

```
    "extra": {
        "zikula": {
            "class": "Zikula\\MenuModule\\ZikulaMenuModule",
            "core-compatibility": ">=3.0.0",
            "displayname": "Menu Module",
            "capabilities": {
                "admin": {
                    "route": "zikulamenumodule_menu_list",
                    "icon": "fas fa-compass"
                }
            },
            "url": "menu",
            "securityschema": {
                "ZikulaMenuModule::": "::",
                "ZikulaMenuModule::id": "::id"
            }
        }
    }
```

Proposed:

```
    "extra": {
        "zikula": {
            "class": "Zikula\\MenuModule\\ZikulaMenuModule",
            "core-compatibility": ">=3.0.0",
            "displayname": "Menu Module",
            "icon": "fas fa-compass",
            "capabilities": {
                "admin": {
                    "route": "zikulamenumodule_menu_list"
                }
            },
            "url": "menu",
            "securityschema": {
                "ZikulaMenuModule::": "::",
                "ZikulaMenuModule::id": "::id"
            }
        }
    }
```

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | fixes #4116 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description

This PR moves the `icon` field out of the `capabilities` array.
Also it does some minor cleanups at composer files, like proper ordering of extra fields and minimum PHP version.
It also updates the VA modules which have these changes already commited.

### DIFF
--- a/src/system/BlocksModule/composer.json
+++ b/src/system/BlocksModule/composer.json
@@ -14,7 +14,7 @@
         "psr-4": { "Zikula\\BlocksModule\\": "" }
     },
     "require": {
-        "php": ">=7.2.0"
+        "php": ">=7.2.5"
     },
     "suggest": {
         "Scribite:>=6.0.2": "To allow WYSIWYG HTML block editing."
@@ -25,23 +25,18 @@
             "core-compatibility": ">=3.0.0",
             "displayname": "Blocks Module",
             "url": "blocks",
-            "oldnames": [],
+            "icon": "fas fa-table",
             "capabilities": {
                 "admin": {
-                    "route": "zikulablocksmodule_admin_view",
-                    "icon": "fas fa-table"
+                    "route": "zikulablocksmodule_admin_view"
                 }
             },
             "securityschema": {
                 "ZikulaBlocksModule::": "Block key:Block title:Block ID",
                 "ZikulaBlocksModule::position": "Position name::Position ID",
-                "ExtendedMenublock::": "Block ID:Link ID:",
                 "fincludeblock::": "Block title::",
                 "HTMLblock::": "Block title::",
                 "HTMLblock::bid": "::bid",
-                "Languageblock::": "Block title::",
-                "Menublock::": "Block title:Link name:",
-                "Menutree:menutreeblock:": "Block ID:Link Name:Link ID",
                 "PendingContent::": "Block title::",
                 "Textblock::": "Block title::",
                 "Textblock::bid": "::bid",
--- a/src/system/ExtensionsModule/Schema/schema.composer.json
+++ b/src/system/ExtensionsModule/Schema/schema.composer.json
@@ -84,7 +84,7 @@
                             "additionalProperties": true
                         }
                     },
-                    "required": ["core-compatibility", "class"],
+                    "required": ["core-compatibility", "class", "icon"],
                     "additionalProperties": false
                 }
             },
--- a/src/system/ExtensionsModule/Schema/schema.composer.json
+++ b/src/system/ExtensionsModule/Schema/schema.composer.json
@@ -73,6 +73,7 @@
                         "class": {"type": "string"},
                         "displayname": {"type": "string"},
                         "url": {"type": "string"},
+                        "icon": {"type": "string"},
                         "oldnames": {"type": "array"},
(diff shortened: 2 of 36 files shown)


src/system/BlocksModule/composer.json
```diff
@@ -25,23 +25,18 @@
             "core-compatibility": ">=3.0.0",
             "displayname": "Blocks Module",
             "url": "blocks",
```
craigh: I don't think `"url"` is valid or used anymore

src/system/ExtensionsModule/Schema/schema.composer.json
```diff
@@ -73,6 +73,7 @@
                         "class": {"type": "string"},
                         "displayname": {"type": "string"},
                         "url": {"type": "string"},
+                        "icon": {"type": "string"},
```
craigh: `"icon"` should be added to the `"required"` list on line 95.

now, I'm not sure about my comment above about `url`. If it is not used and not required, it should be removed here too (line 75)... and then checked in other extensions... but not sure

src/system/ExtensionsModule/Tests/Fixtures/maximum_composer.json
```diff
@@ -38,9 +38,12 @@
       "url": "foo",
       "oldnames": ["food"],
       "capabilities": {
-        "hook_subscriber": {"class": "Zikula\\FooModule\\Container\\HookContainer"},
-        "user": {"route": "zikulafoomodule_foo_index"},
-        "admin": {"route": "zikulafoomodule_bar_index"}
+        "admin": {
```
craigh: `hooks_subscriber` is gone?

src/system/BlocksModule/composer.json
Guite: it is still valid and used

src/system/ExtensionsModule/Tests/Fixtures/maximum_composer.json
Guite: yes, because we do this using service tags now

src/system/ExtensionsModule/Schema/schema.composer.json
Guite: The `required` section is only for 1st level (see it contains `extra` for example), but not for sub elements.

craigh: sorry - I meant line 87

src/system/ExtensionsModule/Tests/Fixtures/maximum_composer.json
craigh: I believe this is still used for `hasCapability` tests, etc.

src/system/ExtensionsModule/Schema/schema.composer.json
Guite: added in 58aabf244ba8fdb03b092e52fb8826eec178a2cf

src/system/ExtensionsModule/Tests/Fixtures/maximum_composer.json
Guite: No. This is all done by collectors.

Guite: See https://github.com/zikula/core/blob/master/src/Zikula/HookBundle/Controller/HookController.php#L68-L75

craigh: that is for hooks but not all capabilities. see https://github.com/zikula/core/blob/master/src/system/ExtensionsModule/Api/CapabilityApi.php#L53

Issue #4161, opened 2020-03-11
"Zikula Native Authorization / Import users" Issue


ThiloWitt:
[TIP]:  # ( Provide a general summary of the issue in the title above ^^ )

| Q                 | A
| ----------------- | ---
| Zikula Version    | 2.0.15
| PHP Version       | 7.3.15

## Expected Behavior
"Zikula Native Authorization / Import users" should import new users 

## Current Behavior
Error Message:
Entity "Zikula\UsersModule\Entity\UserEntity" does not have a setter for property "pass". Please add setPass().
In lib/Zikula/Core/Doctrine/EntityAccess.php (line 148)

            if (method_exists($this, $setMethod)) {
                return $setMethod;
            }
            $class = get_class($this);
            throw new \RuntimeException("Entity \"$class\" does not have a setter for property \"$name\". Please add $setMethod().");
        }
    }

craigh:
the UserEntity doesn't have a `password` (or `pass`) field, so of course it doesn't have a `setPass` method.

craigh:
please provide a copy (or partial copy if very large) of your csv file. obfuscate the passwords if needed

Guite:
Problem seems that https://github.com/zikula/core/blob/master/src/system/ZAuthModule/Helper/FileIOHelper.php#L245 includes `$importValue['pass']`.

Guite:
Also it looks that only the users are created, but not the mapping entities.

craigh:
> Problem seems that https://github.com/zikula/core/blob/master/src/system/ZAuthModule/Helper/FileIOHelper.php#L245 includes $importValue['pass'].

remember, he is using `2.0.15` (you ref'd the master branch)

I think this area was meant to accomplish the mapping by an event maybe?
https://github.com/zikula/core/blob/2.0/src/system/ZAuthModule/Helper/FileIOHelper.php#L274-L278

craigh:
I'll merge this fix with #3265

ThiloWitt:
> 
> 
> please provide a copy (or partial copy if very large) of your csv file. obfuscate the passwords if needed

I this still needed? It is difficult to share the list as these are private data

craigh:
> I this still needed?

no. thank you

ThiloWitt:
Can I use this to update my Zikula 2.0.15 instance? If yes, which patches exactily do I have to implement?

craigh:
@ThiloWitt I'm sorry but I don't think this is possible.

PR #4165, opened 2020-03-14
Add UserCreationApi closes #3265

### LINKED ISSUE #3265: Create UserAdministrationApi
Create a public service in the UsersModule that provides methods to validate and create a new user, delete a user, ...?

refactor Users and ZAuth to use the methods

the create/validate should take an associative array as a single argument.

craigh:
refs #4161 (TBD)

Add a new Api to create users.

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -213,7 +213,7 @@
   - Added "Connections" menu to ExtensionsMenu so extensions can add menu children to other connected extension's admin UI.
   - Added @PermissionCheck annotation for use in Controllers. See `Zikula\PermissionsModule\Annotation\PermissionCheck` and examples in Core.
   - Added display of authentication method to user list (#3704).
-  - Added CreateUsersApi to assist with the manual creation of ZAuth-method users (#3265).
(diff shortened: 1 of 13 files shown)


Guite:
Basically this LGTM.

Two thoughts and ideas:

1. I'd prefer naming it `UsersCreationApi` instead of `CreateUsersApi` (substantiation instead of verb).

2. I find it confusing/wrong that UsersModule implements an interface of ZAuthModule. I'd rather expect the opposite (because ZAuth and other auth modules are adding auth data for a user - so UsersModule should be "leading" here).

Imaging something like:

- UsersModule: UsersCreationApiInterface
- UsersModule: UsersCreationApi implements UsersCreationApiInterface
- UsersModule: UserAuthenticationCreationApiInterface
- ZAuthModule: ZAuthCreationApi implements UserAuthenticationCreationApiInterface
- UsersCreationApiInterface gets/expects an instance of UserAuthenticationCreationApiInterface

Not sure yet if this is 100% perfect. But the idea is to allow creating a user and a selected authentication method's data.

WDYT?

PR #4166, opened 2020-03-15
use global form theme configuration

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description

This PR utilises the `twig.form_themes` configuration option to define a site-wide default for which form themes should be used.

This has two major benefits:

- less redundant code in all the templates
- easier to change at a central place (if for example somebody wants to use Bootstrap 3 or Foundation or a custom style)

Notes:

- To _add_ theme file(s) for a form: `{% form_theme form with ['other.html.twig'] %}`
- To _replace_ theme file(s) for a form: `{% form_theme form with ['other.html.twig'] only %}` 

## Todo Post-Merge

- [ ] VA modules
- [ ] Generator


src/system/CategoriesModule/Resources/views/Form/filter_layout_customizations.html.twig
```diff
@@ -1,10 +0,0 @@
-{% extends 'bootstrap_4_horizontal_layout.html.twig' %}
-{# this template is used in the blocks module #}
-
-{% block form_label_class -%}
-    col-md-3
-{%- endblock form_label_class %}
-
-{% block form_group_class -%}
-    col-md-9
-{%- endblock form_group_class %}
```
craigh: this appears to be completely removed and not replaced?

src/system/RoutesModule/Resources/views/Config/config.html.twig
```diff
@@ -7,7 +7,7 @@
         {% form_theme form with [
             '@ZikulaRoutesModule/Form/bootstrap_4.html.twig',
             '@ZikulaFormExtension/Form/form_div_layout.html.twig'
-        ] %}
+        ] only %}
```
craigh: why isn't Routes using same as all the others?

Guite: Because MOST modules have their own template (inheriting from the default one) because they often declare additional form field types.

src/system/CategoriesModule/Resources/views/Form/filter_layout_customizations.html.twig
Guite: yes; not needed anymore, because it is identical to https://github.com/zikula/core/blob/master/src/Zikula/FormExtensionBundle/Resources/views/Form/bootstrap_4_zikula_admin_layout.html.twig#L3-L9

PR #4191, opened 2020-03-29
Setup Mailer on install

### LINKED ISSUE #3052: Set up mail service on install

craigh:
closes #3052

### DIFF
--- a/src/Zikula/CoreInstallerBundle/Resources/views/Install/mailer.html.twig
+++ b/src/Zikula/CoreInstallerBundle/Resources/views/Install/mailer.html.twig
@@ -1,0 +1,57 @@
+{% extends '@ZikulaCoreInstaller/theme.html.twig' %}
+{% block title 'Zikula Core Installer :: Enter mailer transport information' %}
+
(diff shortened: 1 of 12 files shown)


src/Zikula/CoreInstallerBundle/Resources/views/Install/mailer.html.twig
```diff
@@ -0,0 +1,100 @@
...
+        {{ form_end(form) }}
+    </div>
+</div>
+<script>
+    $(document).ready(function() {
+        $('form:first *:input:text:first').focus();
+
+        $('#submit_button').closest('form').submit(function(event) {
+            $('#db-check').removeClass('d-none');
+        });
```
Guite: This can be removed I think.

Issue #2915, opened 2016-07-10
UsersModule/ZAuthModule improvements


craigh:
- ~implement auto-register setting (see todo in AccessController) to allow attempts to _login_ to auto-create a registration~
- [x] send notification to pending registrations that are deleted (todo in UserAdministrationController::deleteAction)
- [x] ZAuth: make list amount display configurable (currently hardcoded to 25 in UserAdministrationController)
- ~Add ability to 'pair' (or merge) two or more accounts so that authenticating via e.g. Github can point to the same ZikulaUser as authenticating via email or any other method.~ (moved to #4197)

rallek:
User should be able to delete himself. This would be a nice feature as well.

Guite:
Dup of #1322

PR #4192, opened 2020-03-29
Delete Stale registrations

### LINKED ISSUE #4069: Add the ability to delete stale pending registrations
This should also be something the Core does automatically if the time to register expires.

| Q                 | A
| ----------------- | ---
| Zikula Version    | 2.0.15
| PHP Version       | all

## Summary
Deleting all users should be a simple sql call to the database and implemented in UserRegistry. Since they are pending users, they should have no attachment to any other modules so removing them should not cause any consequences.

## Desired Behavior
[NOTE]: # ( Tell us how the new feature should work, be specific )
Have a button in the interface for users would be the simplest implementation. Pressing it gets rid of the users.

## Possible Solution
[NOTE]: # ( Not required, but suggest ideas on how to implement the addition or change )
After pressing it, an Are you Sure dialog would be nice. If yes, all stale pending users are deleted.

## Context
[NOTE]: # ( Why does this feature matter to you? What unique circumstances do you have? )
My sites accumulate a ton of pending users. It woudl be nice to get rid of them.

craigh:
delete stale pending users closes #4069.
Add email notification when registration is deleted refs #2915.

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #4069
| Refs tickets      | #2915
| License           | MIT
| Changelog updated | yes


src/system/RoutesModule/Listener/Base/AbstractUserRegistrationListener.php
```diff
@@ -219,26 +220,26 @@ public function create(GenericEvent $event): void
...
+     *     `echo 'UID: ' . $event->getUser()->getUid();`
      *
      */
-    public function delete(GenericEvent $event): void
+    public function delete(DeletedRegistrationEvent $event): void
     {
     }
-    
+
     /**
```
craigh: @Guite note changes to Generator here also.

PR #4217, opened 2020-04-04
remove extends Event from several events

craigh:
- `Zikula\ExtensionsModule\Event\ConnectionsMenuEvent` no longer extends `Symfony\Contracts\EventDispatcher\Event`.
    - `Zikula\MenuModule\Event\ConfigureMenuEvent` no longer extends `Symfony\Contracts\EventDispatcher\Event`.
    - `Zikula\ThemeModule\Bridge\Event\TwigPostRenderEvent` no longer extends `Symfony\Contracts\EventDispatcher\Event`.
    - `Zikula\ThemeModule\Bridge\Event\TwigPreRenderEvent` no longer extends `Symfony\Contracts\EventDispatcher\Event`.
    - `Zikula\Bundle\FormExtensionBundle\Event\FormTypeChoiceEvent` no longer extends `Symfony\Contracts\EventDispatcher\Event`.
    - `Zikula\Bundle\HookBundle\Hook\Hook` (and all its subclasses) no longer extends `Symfony\Contracts\EventDispatcher\Event`.

### DIFF
--- a/src/Zikula/HookBundle/Twig/Extension/HookExtension.php
+++ b/src/Zikula/HookBundle/Twig/Extension/HookExtension.php
@@ -106,8 +106,11 @@ class HookExtension extends AbstractExtension
      */
     public function notifyFilters(string $content, string $filterEventName)
     {
-        $hook = new FilterHook($content);
+        $hook = $this->hookDispatcher->dispatch($filterEventName, new FilterHook($content));
+        if ($hook instanceof FilterHook) {
+            return $hook->getData();
+        }
 
-        return $this->hookDispatcher->dispatch($filterEventName, $hook)->getData();
+        return $content;
     }
 }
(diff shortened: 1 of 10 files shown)


src/Zikula/HookBundle/Twig/Extension/HookExtension.php
```diff
@@ -106,8 +106,11 @@ public function createRouteUrl(string $routeName, array $routeParameters = [], s
      */
     public function notifyFilters(string $content, string $filterEventName)
     {
-        $hook = new FilterHook($content);
+        $hook = $this->hookDispatcher->dispatch($filterEventName, new FilterHook($content));
```
Guite, line 109: Shouldn't the argument order be swapped?

craigh: The hook dispatcher still takes the arguments in this order for the time being. [It swaps them when dispatching through the eventDispatcher](https://github.com/zikula/core/blob/0c4de2309938827f816cb9e51e50b5b6443decb9/src/Zikula/HookBundle/Dispatcher/HookDispatcher.php#L51). We could consider changing this later.

PR #4232, opened 2020-04-10
remove references to non-existent foundation.

### LINKED ISSUE #3747: Remove Foundation references
We should consider removing all references to the so called "zikula foundation" as it is unclear if the organization ever truly existed. If we can find any documentation instantiating such an organization and its continued existence, then we should add this information to appropriate websites and or repositories. If not, requiring CLA agreements to such an organization is irresponsible at best and illegal at worst.

The question then becomes, what (if anything) is this to be replaced? "Zikula Project" ? Should this have any meaning? Research into other significant organizations would be appropriate.

craigh:
closes #3747


craigh:
styleci failure is a false-fail because the config would need to be changed there first.

PR #4234, opened 2020-04-10
Orphaned users

### LINKED ISSUE #4218: zauth_authentication_mapping not in sync with user table
The zauth_atuthenticaion_mapping table will keep users in it that are unverified, but they get lost from the users table. This causes two results. One, said user can't log in. Two, the route /zauth/admin/list breaks.

| Q                 | A
| ----------------- | ---
| Zikula Version    | 2.0.15
| PHP Version       | 7.3.5

## Expected Behavior

The zauth_atuthenticaion_mapping and users table should always be in sync

## Current Behavior

I am not sure how, but these two tables come out of sync. I think it has to do with verification because the failure only occurs for users that have not verified. In the below image, Ahaddad will show up in the user table and work just fine, but jdewitte, lmigliano, yarrington, and lrkoehler will not.

image

## Possible Solution
After doing some digging I found another bug that may be related to this. If I try to pull up the admin interface to the ZikukaNativeAuth it crashes with an error at Line 119 of AdministrationActionsHelper because this call is returning null.
$userEntity = $this->userRepository->find($mapping->getUid());

You can fix this error by checking for null.
if($userEntity != null) {do stuff}
However, that does not fix the issue. I think I brought this up before, but I could not find the related bug.

## Steps to Reproduce
I am not sure how to reproduce this.

craigh:
fix orphaned users on attempt to delete stale pending registrations.
closes #4218
refs #4069 #4192

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | #4218
| Refs tickets      | #4069, #4192
| License           | MIT


src/system/ZAuthModule/Controller/AccountController.php
```diff
@@ -350,7 +350,6 @@ public function confirmChangedEmailAction(
             return $this->redirectToRoute('zikulausersmodule_account_menu');
         }
         $emailExpireDays = $this->getVar(ZAuthConstant::MODVAR_EXPIRE_DAYS_CHANGE_EMAIL, ZAuthConstant::DEFAULT_EXPIRE_DAYS_CHANGE_EMAIL);
-        $userVerificationRepository->purgeExpiredRecords($emailExpireDays, ZAuthConstant::VERIFYCHGTYPE_PWD, false);
```
Guite, line 353: I can't judge whether this makes sense or not. You decide.

src/system/ZAuthModule/Entity/Repository/UserVerificationRepository.php
```diff
@@ -78,16 +77,18 @@ public function purgeExpiredRecords(
...
-                // delete user record
-                /** @var UserRepositoryInterface $userRepo */
-                $userRepo = $this->_em->getRepository('ZikulaUsersModule:UserEntity');
                 if ($deleteUserEntities) {
-                    /** @var UserEntity $user */
+                    // delete user
                     $user = $userRepo->find($staleVerificationRecord['uid']);
                     $deletedUsers[] = $user;
-                    $userRepo->removeAndFlush($user);
+                    $this->_em->remove($user);
```
Guite, line 88: Maybe:
```
if (null !== $user) {
    $deletedUsers[] = $user;
    $this->_em->remove($user);
}
```
???

src/system/ZAuthModule/Entity/Repository/UserVerificationRepository.php
```diff
@@ -78,16 +77,18 @@ public function purgeExpiredRecords(
...
+                    // delete mapping
+                    $mapping = $authRepo->findOneBy(['uid' => $staleVerificationRecord['uid']]);
+                    $this->_em->remove($mapping);
```
Guite, line 91: Maybe:
```
if (null !== $mapping) {
    $this->_em->remove($mapping);
}
```
???

src/system/ZAuthModule/Helper/LostPasswordVerificationHelper.php
```diff
@@ -121,9 +121,6 @@ public function decodeLostPasswordId(string $identifier = ''): array
      */
     public function checkConfirmationCode(int $userId, string $code): bool
     {
-        $changePasswordExpireDays = $this->variableApi->get('ZikulaZAuthModule', ZAuthConstant::MODVAR_EXPIRE_DAYS_CHANGE_PASSWORD, ZAuthConstant::DEFAULT_EXPIRE_DAYS_CHANGE_PASSWORD);
-        $this->userVerificationRepository->purgeExpiredRecords($changePasswordExpireDays, ZAuthConstant::VERIFYCHGTYPE_PWD, false);
```
Guite, line 125: I can't judge whether this makes sense or not. You decide.

PR #4242, opened 2020-04-12
make AssetBag 'dumb' again.

craigh:
make AssetBag 'dumb' again. move checks to resolver class. update pageAddAsset() calls to use weight for jquery-ui. resolvers don't need full kernel, only env. update tests to use assertSame for arrays.

### DIFF
--- a/src/Zikula/FormExtensionBundle/Resources/views/Form/bootstrap_4_zikula_admin_layout.html.twig
+++ b/src/Zikula/FormExtensionBundle/Resources/views/Form/bootstrap_4_zikula_admin_layout.html.twig
@@ -158,7 +158,7 @@ col-md-9
     {% endif %}
     {# load required assets: jQuery UI + custom search code #}
     {{ pageAddAsset('stylesheet', asset('jquery-ui/themes/base/jquery-ui.min.css')) }}
-    {{ pageAddAsset('javascript', asset('jquery-ui/jquery-ui.min.js')) }}
+    {{ pageAddAsset('javascript', asset('jquery-ui/jquery-ui.min.js'), constant('Zikula\\ThemeModule\\Engine\\AssetBag::WEIGHT_JQUERY_UI')) }}
     {{ pageAddAsset('stylesheet', zasset('@ZikulaUsersModule:css/livesearch.css')) }}
     {{ pageAddAsset('javascript', zasset('@ZikulaUsersModule:js/Zikula.Users.LiveSearch.js')) }}
 {% endblock %}
--- a/src/system/ThemeModule/Engine/Asset/JsResolver.php
+++ b/src/system/ThemeModule/Engine/Asset/JsResolver.php
@@ -76,7 +76,7 @@ class JsResolver implements ResolverInterface
             if ('jquery-ui.min.js' === mb_substr($source, -16)
                 || 'jquery-ui.js' === mb_substr($source, -12)
             ) {
-                $this->bag->add([$source => AssetBag::WEIGHT_JQUERY]);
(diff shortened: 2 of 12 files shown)


src/system/ThemeModule/Engine/Asset/JsResolver.php
```diff
@@ -68,4 +67,17 @@ public function getBag(): AssetBag
...
+
+    private function validateBag(): void
+    {
+        foreach ($this->getBag()->all() as $source) {
+            // if already there, add jquery-ui again to force weight setting is correct
+            // jQueryUI must be loaded before Bootstrap, refs #3912
+            if ('jquery-ui.min.js' === mb_substr($source, -16)
+                || 'jquery-ui.js' === mb_substr($source, -12)
+            ) {
+                $this->bag->add([$source => AssetBag::WEIGHT_JQUERY]);
```
Guite, line 79: Why not `WEIGHT_JQUERY_UI` here?

craigh: fixed

src/Zikula/FormExtensionBundle/Resources/views/Form/bootstrap_4_zikula_admin_layout.html.twig
```diff
@@ -158,7 +158,7 @@ col-md-9
     {% endif %}
     {# load required assets: jQuery UI + custom search code #}
     {{ pageAddAsset('stylesheet', asset('jquery-ui/themes/base/jquery-ui.min.css')) }}
-    {{ pageAddAsset('javascript', asset('jquery-ui/jquery-ui.min.js')) }}
+    {{ pageAddAsset('javascript', asset('jquery-ui/jquery-ui.min.js'), constant('Zikula\\ThemeModule\\Engine\\AssetBag::WEIGHT_JQUERY_UI')) }}
```
Guite, line 161: Why are these needed? Shouldn't this be done automatically (in/by `validateBag`)?
Is this just for verbosity or is it required to add the weight explicitly for each `pageAddAsset` call now?

craigh: it is not required (for jquery-ui only) of course, but frankly, I would prefer that it is set manually always and remove from validateBag entirely.

src/system/ThemeModule/Engine/Asset/JsResolver.php
```diff
@@ -39,19 +38,19 @@ class JsResolver implements ResolverInterface
...
+        string $env,
         AssetBag $bag,
         MergerInterface $merger,
         bool $combine = false
     ) {
-        $this->kernel = $kernel;
         $this->bag = $bag;
         $this->merger = $merger;
-        $this->combine = 'prod' === $this->kernel->getEnvironment() && $combine;
+        $this->combine = ('prod' === $env) && $combine;
```
Guite, line 48: cs: please remove the unneeded parenthesis again

craigh: in this case I prefer them.  a === b && c is confusing to look at.

src/Zikula/FormExtensionBundle/Resources/views/Form/bootstrap_4_zikula_admin_layout.html.twig
Guite: > remove from validateBag entirely

please let's keep it until 4.0 when we finally switch to modern JS handling (ES6, webpack, etc.)

Issue #4197, opened 2020-03-31
[ZAuth/OAuth] Merge accounts with same email


craigh:
Transferred from #2915

Add ability to 'pair' (or merge) two or more accounts so that authenticating via e.g. Github can point to the same ZikulaUser as authenticating via email or any other method.

refs #3720

craigh:
basically - assuming duplicate emails are **_not allowed_**, the process would be this:

when someone logs in with ZAuth or OAuth, it would check emails in users table and see if the email exists. Assuming the authentication succeeds (verifying then that they indeed own the email address), the same UID would be used as the existing email record, instead of creating a new one.

Taking care of existing duplicate accounts would be more challenging. A couple possibilities I can think of:

1. allow a user to manually merge accounts as they see fit
2. on upgrade scan for duplicate emails in mapping tables and users table and merge the accounts.

second option would be easier, but there is possibility of data loss. For example a comment may  be created with the 'lost' UID. This comment would be orphaned.

robbrandt:
I've thought about this for a little while and I am not certain this is a problem that needs to be solved.

I am using alternative authentication methods on 3 different sites, probably 4 soon. In my situation, duplicate emails and usernames aren't normal but are common for administrators. The alternate authentication method is the primary means of authentication, but ZAuth is always there as a fallback.

To me, the biggest issue is identifying **which** username uses which auth method, when trying to edit them or assign them to groups.  I think I saw a solution for this posted or at least suggested?

If we are too strict, the following scenario is problematic: user registers initially as rob.brandt using zuath.  Site administrator then adds support for a Facebook login, and naturally rob.brandt wants to start using that.  But it won't work because his existing Facebook identity is rob.brandt, and the **same email address** as used with zauth.

Kaik:
@robbrandt I think this is what Craigh is talking about if there is an account with the same email then this account is linked to fb user, no new account is created. (I don't think zk account was created before anyway... remember comment in the code "// create new account here"...)

@craigh 
while option number 1 will leave duplicates untill these are merged by user

Merging accounts will be/is quite a big problem, maybe creating an event would be a good solution so modules can respond same way as in case of deleting an user... sometimes some users want to merge their different accounts into one so it would be helpful as well.

craigh:
this issue must be resolved with #3720

craigh:
[...]

PR #4271, opened 2020-04-22
restrict ability to register same email with different auth methods

### LINKED ISSUE #3720: Duplicate emails are allowed
| Q                 | A
| ----------------- | ---
| Zikula Version    | 1.5.0 ci 248
| PHP Version       | 5.6

### Expected behavior
If an email address is allready registered another usershould not be able to use the same one. This is also valid even if the user is registering with different authentication methods

### Actual behavior
a user (me) is registered with an email via ZAuth. With the same email address I am registered at facebook. If I now register via facebook/oauth it is possible without any problem. I looked into the database and can see both users do have the same email address

### Steps to reproduce
see actual behavior

Is it a bug or is it a feature?

craigh:
closes #3720
refs #4197

### DIFF
--- a/src/system/UsersModule/Controller/RegistrationController.php
+++ b/src/system/UsersModule/Controller/RegistrationController.php
@@ -20,6 +20,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+use Symfony\Component\Validator\ConstraintViolationInterface;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 use Zikula\Bundle\CoreBundle\Controller\AbstractController;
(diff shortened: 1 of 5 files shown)


src/system/UsersModule/Controller/RegistrationController.php
```diff
@@ -161,11 +163,21 @@ public function registerAction(
...
+                        /** @var ConstraintViolationInterface $validationError */
                         foreach ($validationErrors as $validationError) {
                             $this->addFlash('error', $validationError->getMessage());
+                            $codes[] = $validationError->getCode();
+                        }
+                        if ($authenticationMethod instanceof ReEntrantAuthenticationMethodInterface) {
+                            $session->remove('authenticationMethod');
+                        }
+                        $route = 'zikulausersmodule_registration_register';
+                        if (in_array(ValidUserFieldsValidator::DUP_EMAIL_ALT_AUTH, $codes)) {
```
Guite, line 176: Please add `true` as third parameter.

src/system/UsersModule/Entity/Repository/UserRepository.php
```diff
@@ -217,4 +217,35 @@ public function searchActiveUser(array $unameFilter = [])
 
         return $this->query($filter, ['uname' => 'asc']);
     }
+
+    public function countDuplicateUnames(string $uname, ?int $uid = null): int
+    {
+        $qb = $this->createQueryBuilder('u');
+        $qb->select('count(u.uid)')
+            ->where($qb->expr()->eq('LOWER(u.uname)', ':uname'))
+            ->setParameter('uname', $uname);
```
Guite, line 226: `strtolower($uname)`?

Guite:
`MockUserRepository` needs to implement the two new methods (see Travis).

PR #4274, opened 2020-04-23
allow users to self delete with admin setting

### LINKED ISSUE #1322: Allow user to delete own account
Feature Request and Discussion

In "My Account" a user can change their email and password. Similarly they should be able to delete their user account.
## Issues
- What happens to data throughout the site that is tagged with that users username or ID? I would think that at the least that data should no longer show their username.
- When a user deletes their account should they also be given the chance to delete items they have contributed such as forum posts/topics, uploads etc.?
- Should the username that they delete be immediately available for a new user to register?
- Others?

craigh:
fixes #1322


Guite:
Nice that ghosted user names are added to list of illegal user names.

craigh:
what happened to Travis?

PR #4281, opened 2020-04-27
reset selected modules to version 2.9.9 on upgrade 

### LINKED ISSUE #4280: sync core extension version numbers with core release version
this will ease development

refs #3967

already done some here https://github.com/zikula/core/commit/8ddc500348db78f3d6b2cccd19007dbd1a911fba

Still to do:
- [ ] blocks
- [ ] extensions
- [ ] theme
- [ ] users
- [ ] Bundles?

craigh:
and run needed upgrades for 3.0.0
fixes #4280 

remove dead code from upgraders since we do not support upgrades from before Core-1.4.3.
remove unneeded code comments


src/system/BlocksModule/BlocksModuleInstaller.php
```diff
@@ -20,167 +20,48 @@
...
         BlockPositionEntity::class,
         BlockPlacementEntity::class
     ];
 
-    /**
-     * @var ZikulaHttpKernelInterface
-     */
-    private $kernel;
-
     public function __construct(
```
Guite, line 39: You can remove the whole constructor now.

Issue #4284, opened 2020-05-01
display issues


craigh:
- [x] at `/blocks/admin/block/edit/1` (click 'add filter') - larger form elements have unwanted `custom-select` class. Think this causes them to ignore `form-control-sm` class.

![Screen Shot 2020-04-30 at 7 23 59 PM](https://user-images.githubusercontent.com/350048/80804352-dd1e9600-8b82-11ea-977f-ea186c310dd5.png)

 - [ ] at `/menu/admin/view/1` - missing translation or label (e.g. "null")

![Screen Shot 2020-05-01 at 8 05 12 AM](https://user-images.githubusercontent.com/350048/80804353-ddb72c80-8b82-11ea-8167-af5ca24d5acf.png)

Guite:
> missing translation or label (e.g. "null")

This is a general problem with JS translations: if no translation value is set it does not use the translation key, but treats the `null` value as translation value.

Guite:
> larger form elements have unwanted `custom-select` class. Think this causes them to ignore `form-control-sm` class

Use: `custom-select-sm` (see [BS4 docs](https://getbootstrap.com/docs/4.4/components/forms/#select-menu)).

PR #4287, opened 2020-05-02
update translations

craigh:
this is a result of first running

```shell
bin/console translation:extract zikula en
```

then running

```shell
bin/console translation:update en --output-format=yaml --sort --force --prefix=''
```

refs #4284


craigh:
I've no problem with closing this unmarked, I just wanted you to see the result of what I learned.

PR #4297, opened 2020-05-08
Symfony insight fixes

craigh:
fixes for symfony insight concerns

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -28,9 +28,7 @@ class Kernel extends ZikulaKernel
 
     public function registerBundles(): iterable
     {
-        $databaseUrl = $this->booted && $this->getContainer()->hasParameter('database_dsn') ? $this->getContainer()->getParameter('database_dsn') : '';
-        $bundleHelper = new PersistedBundleHelper($databaseUrl);
-//        $bundleHelper = new PersistedBundleHelper($_ENV['DATABASE_URL'] ?? '');
+        $bundleHelper = new PersistedBundleHelper($_ENV['DATABASE_URL'] ?? '');
         $bundles = require $this->getProjectDir() . '/config/bundles.php';
         $bundleHelper->getPersistedBundles($this, $bundles);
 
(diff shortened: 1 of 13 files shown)


src/Kernel.php
```diff
@@ -28,7 +28,9 @@ class Kernel extends ZikulaKernel
 
     public function registerBundles(): iterable
     {
-        $bundleHelper = new PersistedBundleHelper($_ENV['DATABASE_URL'] ?? '');
+        $databaseUrl = $this->booted && $this->getContainer()->hasParameter('database_url') ? $this->getContainer()->getParameter('database_url') : '';
+        $bundleHelper = new PersistedBundleHelper($databaseUrl);
+//        $bundleHelper = new PersistedBundleHelper($_ENV['DATABASE_URL'] ?? '');
         $bundles = require $this->getProjectDir() . '/config/bundles.php';
```
craigh, line 34: @Guite this is an idea but I've not really tested it. it requires the parameter change in services.yaml too. I'm not sure its better practice then using the GLOBAL. :-) wonder what you think.

craigh: actually, I'm not even sure it WORKS. I'll be interested to hear your thoughts.

Guite: Yes. Using a parameter seems the way to go :+1:

src/system/ExtensionsModule/Twig/Extension/ExtensionsExtension.php
```diff
@@ -19,16 +19,6 @@
 
 class ExtensionsExtension extends AbstractExtension
 {
-    /**
-     * @var TranslatorInterface
-     */
-    private $translator;
-
-    public function __construct(TranslatorInterface $translator)
```
Guite, line 27: Please remove `use Symfony\Contracts\Translation\TranslatorInterface;` too (see StyleCI).

src/Kernel.php
craigh: doesn't work.

Issue #4095, opened 2020-01-30
Changes for Symfony 5.1


Guite:
This issue collect some changes to be done when Symfony is upgraded to 5.1

- [x] Upgrade (done in #4319)
- [x] [Constants for command exit codes](https://symfony.com/blog/new-in-symfony-5-1-misc-improvements-part-1) (done in #4321)

craigh:
something we could do here is require symfony dev-master branch (assuming it is stable enough to use) and implement all these changes now. Then when 5.1 is released we won't have to do anything except change the composer requirement.

Guite:
No need. We have other tasks on the list and 5.1 isn't far away.

Guite:
> use compound constraints

Where? ZAuth? Users?

craigh:
> Remove `secret`

FYI - the secret is also migrated on upgrade so this should be accommodated.

craigh:
I don't think we can entirely eliminate `bootstrap.php` because we have some [custom stuff](https://github.com/zikula/core/blob/master/config/bootstrap.php#L41-L45) happening there

craigh:
![Screen Shot 2020-05-04 at 7 57 33 AM](https://user-images.githubusercontent.com/350048/80963437-ec008500-8ddc-11ea-96eb-497f051824ce.png)

PR #4319, opened 2020-05-31
Update to Symfony 5.1.0

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | yes (by vendor)
| Fixed tickets     | -
| Refs tickets      | #4095
| License           | MIT
| Changelog updated | yes

## Description

- Update/simplify `bootstrap.php` wrt `DotEnv` constructor
- Use new `DotEnv::bootEnv` method
- Remove `secret` parameter from `config/packages/framework.yaml`
- Update/simplify `Kernel` class utilizing [improved Microkernel](https://symfony.com/blog/new-in-symfony-5-1-improved-microkernel)

### DIFF
--- a/config/packages/framework.yaml
+++ b/config/packages/framework.yaml
@@ -1,5 +1,4 @@
 framework:
-    secret: '%env(APP_SECRET)%'
     csrf_protection: ~
     #http_method_override: true
 
@@ -22,3 +21,8 @@ framework:
     #fragments: true
     php_errors:
         log: true
+
+parameters:
+    # settings for container dumper
+    container.dumper.inline_factories: true
+    container.dumper.inline_class_loader: false
--- a/src/system/SecurityCenterModule/Listener/FilterListener.php
+++ b/src/system/SecurityCenterModule/Listener/FilterListener.php
@@ -32,6 +32,7 @@ use Symfony\Component\Mime\Address;
 use Symfony\Component\Mime\Email;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Symfony\Contracts\Translation\TranslatorInterface;
+use Zikula\Bundle\CoreBundle\HttpKernel\ZikulaHttpKernelInterface;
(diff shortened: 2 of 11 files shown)


src/system/SecurityCenterModule/Listener/FilterListener.php
```diff
@@ -224,6 +232,7 @@ public function idsInputFilter(RequestEvent $event): void
      */
     private function getIdsConfig(): array
     {
+        $vendorDir = $this->kernel->getProjectDir() . '/vendor/';
```
craigh, line 235: probably could inject only the `project_dir` instead of the full kernel.

config/packages/framework.yaml
```diff
@@ -1,5 +1,4 @@
 framework:
-    secret: '%env(APP_SECRET)%'
```
craigh, line 2: I think the core-upgrade/install process does something with the secret? can't remember.

Guite: It only takes it from old installations to migrate it into the `env` file. If the parameter does not exist it generates a new secret instead. So no problem to remove it for new installs.

src/system/SecurityCenterModule/Listener/FilterListener.php
Guite: Which syntax would that use?

craigh: there are many examples in the core

craigh: https://github.com/zikula/core/blob/50cacd745ac12e9c81bedda7a6193e130194b02b/src/Zikula/CoreBundle/Resources/config/services.yaml#L8

Guite: updated

PR #4326, opened 2020-06-06
corrections to add listeners to hooks

craigh:
previous implementation required hook providers use public services. this does not.


Guite:
:+1:

PR #4369, opened 2020-07-02
Add StaticContentModule

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes!
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description

~Added template display controller for quick generation and display of static template-based content.~

## To do
 ~- [x] docs~
 ~- [x] also add template-based block~
 ~- [x] address question of required `/p/*` in route~
 ~- [ ] add permission scheme somewhere~

- move content-blocks to new StaticContentModule
- update docs
- update maker (`make:zikula-extension`) to discourage use of Zikula as vendor
~- [ ] update BlocksModuleInstaller to shift new block creation on install to StaticContentModule and update existing classes.~


craigh:
it is quite simple really:

1. add a template like `/templates/p/foo.html.twig`
2. call `/p/foo`

Guite:
Nice!

Just add docs please and it should be ready.

Guite:
> provide the permission scheme

means `composer.json` meta data

craigh:
> Any idea where to provide the permission scheme

yes, I wondered about that also. I didn't have any good ideas. ExtensionsModule seems wrong unless the controller is moved there. But the controller doesn't belong there.

craigh:
The scheme is not required for operation. It is only informational.

Guite:
How should an admin know about it if it is not populated via composer?
If the admin can not know about it it becomes useless because no permission rules will be created for it.

craigh:
> How should an admin know about it if it is not populated via composer?

of course. I'm not suggesting it should be eliminated. I'm just stating that the composer information doesn't affect functionality at all.

maybe we should reconsider this information collection for a later release.

In any case, I'm not certain the best way to handle this currently.

Guite:
> maybe we should reconsider this information collection for a later release.

agreed; it could also be populated programmatically (`$someService->populatePermissionScheme()`)

craigh:
I also got thinking about the route. currently this PR requires `/p/{name}`. I did this to easily avoid route conflicts with other modules and made it one letter to keep it simple.

My personal use-case for this feature is to show a Google calendar. Google provides the html, I paste it in a template and can show it at `/p/calendar` .. but now I am thinking I would rather that be simply `/calendar` because when I am talking to someone and say "go to foo.com slash calendar" it is better than "go to foo.com slash P slash calendar".

I think simply having the route be `/{name}` would be a problem though.
Thoughts?

Guite:
I think this would be a use case for adding a custom route (Routes module) to let `calendar` point to the same controller as `p/calendar` (additionally).
Centrally adding `/{name}` would certainly lead to conflicts.

craigh:
> I think this would be a use case for adding a custom route (Routes module)

this is currently impossible (as far as I can tell):

Screen Shot 2020-07-06 at 9 24 24 AM

1. The bundle is not listed (because it does not include CoreBundle)
2. There is no 'action' in this case as it is a single-action-controller.

An idea I had to solve the permission scheme issue and the bundle not being listed (1 above) is to move all this to the Routes module. This makes some sense to me because we are adding a special type of route to display templates... not 100% about SOC here though.

for issue 2 above, we would have to either change the controller to include an action or change the routes module to not require an action - OR remove the `/p` prefix).

what do you think?

Guite:
> The bundle is not listed

This is not a dropdown (like in the Settings module), but a text field. So there is nothing listed, but needs to be entered manually.

> There is no 'action' in this case as it is a single-action-controller

Indeed. This is not supported for custom routes atm.

craigh:
> This is not a dropdown

The **Bundle** selector _is_ a `select` element, listing only extensions (not bundles)

Screen Shot 2020-07-06 at 9 46 15 AM

craigh:
> move all this to the Routes module.

thoughts about this?

Guite:
I think that would be a viable option.

docs/Content/Pages.md
```diff
@@ -0,0 +1,25 @@
...
+ - `{{ app.request.method }}`
+ - `{{ pageAddAsset('javascript', asset('/path/from/public/to/my/javascript.js')) }}`
+ - etc...
+
+There is no way to add additional variables in these pages.
+
+### Better Pages
+To display more advanced pages and non-static content please see
+ - [Pages](https://github.com/zikula-modules/Pages)
+ - [Content](https://github.com/zikula-modules/Content)
```
Guite, line 25: Maybe link to https://docs.ziku.la/Extensions/content.html ?

Guite:
Awesome progress :+1: 

Suggestion for updating existing blocks:

```
UPDATE `blocks`
SET `bkey` = REPLACE(`bkey`, 'BlocksModule', 'StaticContentModule')
WHERE `blocktype` IN ('Finclude', 'Html', 'Text', 'Xslt');
```

craigh:
@Guite I am having second thoughts about removing the contents blocks from Blocks module at 3.1.0. I am thinking full removal should wait until 4.0. In the meantime, I think we should add StaticContent module and deprecate the old blocks with plans to remove at 4.0. In the meantime, we can post an alert to the blocks admin that encourages user to move blocks to StaticContent. Then we don't do anything in the upgrade, etc. thoughts?

Guite:
That would indeed be cleaner. Good idea.

Guite:
terrific :smiley:

PR #4396, opened 2020-07-11
Allow creating and changing mappings without setting a password

### LINKED ISSUE #4395: Creating a new user without password is not possible for the admin
[TIP]:  # ( Provide a general summary of the issue in the title above ^^ )

| Q                 | A
| ----------------- | ---
| Zikula Version    | 3.0.0
| PHP Version       |7.4.5

## Expected Behavior

Creating a new user without password shall be possible for the admin

## Current Behavior
Creating a new user for Zikula Native Authorisation
If the password generation is unchecked you can not save the new user. There is still the password required.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | fixed #4395 
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description

- Corrected wrong field reference in JS to update input field required states depending on the "set password" toggle
- Allow calling the `setPass` method with `null` value, but do not assign these to avoid overriding existing passwords
  - Otherwise submitting the form without having a new password set results in `Expected argument of type "string", "null" given at property path "pass".`

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -23,6 +23,7 @@
   - Improved responsive design of the extensions list.
   - Throw more descriptive exception when non-zasset asset is not found (#4366).
   - Install extension assets on installation and upgrade (#4367).
+  - Allow creating and changing mappings without setting a password  (#4395).
 
 - Features:
   - _there should be none_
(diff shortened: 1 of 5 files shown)


craigh:
I am also getting this JS error when trying to submit the "Create New User" form. Does this PR also fix that?
> An invalid form control with name='zikulazauthmodule_admincreateduser[pass][pass][first]' is not focusable.

It only occurs if "set password now" is disabled, so the password fields are not visible.

Guite:
> Does this PR also fix that?

Likely

craigh:
I checked - it does fix it.

PR #4429, opened 2020-07-28
fix/update StyleCI config

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | depends on the point of view
| New feature?      | depends on the point of view
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description

StyleCI did some updates to their platform and rulesets.

See https://blog.styleci.io/platform-improvements/

This PR updates our configuration accordingly.
Most changes remove individual supplements and additions which became unneeded, since they are now part (or not part) of the `symfony` preset anyway (and StyleCI gives an error if things are repeated).

There are one or two exceptions where things are differently handled now; but they fit the more modern coding style, which we now use with 3.x.

Please look at https://github.styleci.io/analyses/RvDl1D for the changeset that results from the changes.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,57 +1,40 @@
 preset: symfony
+risky: true
 
 enabled:
   - alpha_ordered_imports
   - combine_consecutive_issets
   - combine_consecutive_unsets
-  - compact_nullable_typehint
   - declare_strict_types
-  - dir_constant
-  - ereg_to_preg
   - escape_implicit_backslashes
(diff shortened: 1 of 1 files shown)


craigh:
looks mostly fine. the only [change](https://github.styleci.io/analyses/RvDl1D) that looked odd was to `src/system/UsersModule/Constant.php`

fine with me to merge

Issue #4352, opened 2020-06-18
[General] remove symfony/symfony dependency


craigh:
remove symfony/symfony dependency and replace with only needed components.

this is required as part of https://symfony.com/doc/current/setup/flex.html

at least

but also, just good practice and also current best-practice.

Guite:
isn't this necessarily for 4.0.0 ?
because if we remove something this is a BC break

craigh:
the goal would be to not remove anything 😄 

I believe this simply done.

Guite:
Then the description above is misleading:

> replace with **only needed** components

craigh:
possible you are right. I've done some preliminary research. As far as I can tell, the _core_ is not using

 - component/browser-kit // used in functional tests
 - component/css-selector // used in functional tests
 - component/dom-crawler // used in functional tests
 - component/http-client ?? no? maybe used in functional tests?
 - component/inflector
 - component/ldap
 - component/lock
 - component/process ??
 - component/property-access
 - component/property-info
 - component/serializer
 - component/stopwatch
 - component/templating (maybe used by maker?)
 - component/var-dumper
 - component/var-exporter
 - component/web-link

Guite:
If/After #4514 has been merged the String component's `functions.php` file is included in `bootstrap.php` to make the helper methods available (`u()`, `b()`, `s()`).
This inclusion can be removed again as part of #4352 because then the String component will be explicitly added by composer which adds `"files": [ "Resources/functions.php" ]` to the autoloader ([link](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/String/composer.json#L34)).

PR #4433, opened 2020-08-03
Replace custom parameters with real symfony config

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | yes
| Fixed tickets     | -
| Refs tickets      | #4352
| License           | MIT
| Changelog updated | yes

## Description

The goal of this PR would be to replace all custom parameters in the `services_custom.yaml` file with appropriate configuration in core extensions/bundles using "real" symfony config.

Ultimately, I'd like to remove the `services_custom.yaml` completely since the `services.yaml` file is _entirely custom anyway_. I would also like to look at removing the `dynamic/generated.yaml` and `default` files as well.

This would all need some BC work so that 3.1.0 doesn't break existing installs, but I think this is quite doable. 

## Todos

- [x] Tests
- [x] Documentation
- [x] Changelog


docs/zikula_theme.yaml
```diff
@@ -0,0 +1,7 @@
+zikula_theme:
+    script_position: head
+    asset_manager:
+        combine: false
+        lifetime: '1 day'
+        compress: true
+        minify: true
```
craigh, line 7: this is currently in the /docs dir to show the structure of the package file if needed. This would ultimately need to be part of the docs.

PR #4444, opened 2020-08-11
avoid exception by potential autoloader closure

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description

This PR avoids that additional vendors that contribute own autoloaders may cause a problem in our kernel if these autoloaders are closures and hide (shadow) the autoloader from composer.

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -6,6 +6,7 @@
   - _there should be none_
 
 - Fixes:
+  - [CoreBundle] More robust autoloader detection.
   - [Menu] Fix error during creation of new menus.
 
 - Features:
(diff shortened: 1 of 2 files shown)


craigh:
Is this from a recipe or did you find a problem?

Guite:
I found the problem (exception _"Cannot use object of type Closure as array"_) in an environment where vendors included additional libraries (e.g. from Zend).

craigh:
👍👍

PR #4481, opened 2020-09-18
remove action suffix from controller methods

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | yes
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

## Description

This PR removes the `Action` suffix from controller method names, because this is neither required nor recommended anymore.

From https://symfony.com/doc/4.2/best_practices/controllers.html#controller-action-naming :

> Don’t add the `Action` suffix to the methods of the controller actions.

> The first Symfony versions required that controller method names ended in `Action` (e.g. `newAction()`, `showAction()`). This suffix became optional when annotations were introduced for controllers. In modern Symfony applications this suffix is neither required nor recommended, so you can safely remove it.

### DIFF
--- a/CHANGELOG-3.1.md
+++ b/CHANGELOG-3.1.md
@@ -21,6 +21,6 @@
 
 - Deprecated:
   - [General] Controller methods should not have an `Action` suffix in their names anymore.
-  - [CoreBundle] `Zikula/CoreBundle/YamlDumper` use `Configurator` as needed.
+  - [CoreBundle] `Zikula/CoreBundle/YamlDumper` is deprecated. Please use `Configurator` as needed.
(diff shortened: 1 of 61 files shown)


craigh (changes requested):
please investigate travis failure.

Guite:
> Invalid version string "action-suffix.x-dev"

No idea why composer errors like this. I hoped it would be just a temporary issue. At least it seems to be connected to the branch name instead of the code changes.

craigh (approved):
very well. consider naming future branches differently. :-)

Issue #4376, opened 2020-07-06
[Routes] refactor to allow Single Action Controllers


craigh:
It is currently impossible to implement single action ([invokable](https://symfony.com/doc/current/controller/service.html#invokable-controllers)) controllers because the routing module expects standard stuff.

 - src/system/RoutesModule/Routing/RouteLoader.php:283
 - create a custom route requires the controller and action, etc...

Guite:
Closing as won't fix as we are about to remove the custom routing system.

PR #4514, opened 2020-10-06
first usages of String component

### LINKED ISSUE #4053: [General] Utilise String component
We should use the string component of Symfony 5 to make string operations both more readable and more solid (less prone to errors).

https://symfony.com/blog/new-in-symfony-5-0-string-component
https://symfony.com/doc/current/components/string.html

We should wait until Symfony 5.1 is available though so that we can rely on the API.

Guite:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | fixed #4053 
| Refs tickets      | -
| License           | MIT
| Changelog updated | no

## Description

This PR introduces the String component for some string operations.

Currently the `functions.php` file is explicitly included to make the helper methods available (`u()`, `b()`, `s()`).
This inclusion can be removed again when #4352 is done because then the String component will be explicitly added by composer which adds `"files": [ "Resources/functions.php" ]` to the autoloader ([link](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/String/composer.json#L34)).

### DIFF
--- a/src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
+++ b/src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
@@ -133,7 +133,7 @@ EOT
         $this->verified = in_array((int) $input->getOption('verified'), [0, 1, 2]) ? (int) $input->getOption('verified') : 1;
         $regDate = $input->getOption('regdate') ?? $this->nowUTC;
         $this->range = is_string($regDate) && '>' === $regDate[0];
-        $this->regDate = $this->range ? (string) s($regDate)->slice(1) : $regDate;
+        $this->regDate = $this->range ? s($regDate)->slice(1)->__toString() : $regDate;
 
         $io->title('User generation utility');
         $io->text('Generating users...');
--- a/src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
+++ b/src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
@@ -133,7 +133,7 @@ EOT
         $this->verified = in_array((int) $input->getOption('verified'), [0, 1, 2]) ? (int) $input->getOption('verified') : 1;
         $regDate = $input->getOption('regdate') ?? $this->nowUTC;
         $this->range = is_string($regDate) && '>' === $regDate[0];
-        $this->regDate = $this->range ? s($regDate)->slice(1)->__toString() : $regDate;
+        $this->regDate = $this->range ? s($regDate)->slice(1)->toString() : $regDate;
 
         $io->title('User generation utility');
         $io->text('Generating users...');
--- a/src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
+++ b/src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
@@ -133,7 +133,7 @@ EOT
         $this->verified = in_array((int) $input->getOption('verified'), [0, 1, 2]) ? (int) $input->getOption('verified') : 1;
         $regDate = $input->getOption('regdate') ?? $this->nowUTC;
         $this->range = is_string($regDate) && '>' === $regDate[0];
-        $this->regDate = $this->range ? (string)s($regDate)->slice(1) : $regDate;
+        $this->regDate = $this->range ? (string) s($regDate)->slice(1) : $regDate;
 
         $io->title('User generation utility');
         $io->text('Generating users...');
--- a/src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
+++ b/src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
@@ -15,6 +15,7 @@ namespace Zikula\Bundle\CoreBundle\Command;
 
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
+use function Symfony\Component\String\s;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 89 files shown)


craigh:
> This inclusion can be removed again when #4352 is done because then the String component will be explicitly added by composer

why not add this to our own composer file?

src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
```diff
@@ -132,7 +133,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $this->verified = in_array((int) $input->getOption('verified'), [0, 1, 2]) ? (int) $input->getOption('verified') : 1;
         $regDate = $input->getOption('regdate') ?? $this->nowUTC;
         $this->range = is_string($regDate) && '>' === $regDate[0];
-        $this->regDate = $this->range ? mb_substr($regDate, 1) : $regDate;
+        $this->regDate = $this->range ? (string)s($regDate)->slice(1) : $regDate;
```
craigh, line 136: space after cast or no? here you do not have a space, later you do have a space.

src/Zikula/CoreBundle/Composer/MetaData.php
```diff
@@ -88,7 +89,7 @@ public function __construct(array $json = [])
         $this->dependencies = $this->formatDependencies($json);
         $this->shortName = $json['extra']['zikula']['short-name'];
         $this->class = $json['extra']['zikula']['class'];
-        $this->namespace = mb_substr($this->class, 0, mb_strrpos($this->class, '\\') + 1);
+        $this->namespace = (string) s($this->class)->beforeLast('\\', true);
```
craigh, line 92: space here.

Also, it seems strange to me that a component called **String** requires to be cast as a string... shouldn't it output a string? I know that it is returning an `AbstractSting` that is `\Stringable`, so can we lose the repeated `(string)` casting?

src/Zikula/CoreBundle/Composer/Scanner.php
```diff
@@ -73,18 +74,18 @@ public function decode(string $jsonFilePath)
     {
         $base = str_replace('\\', '/', dirname($jsonFilePath));
         $srcPath = dirname(__DIR__, 4) . '/src/';
-        $base = mb_substr($base, mb_strlen($srcPath));
+        $base = s($base)->slice(mb_strlen($srcPath));
```
craigh, line 77: now here you do not cast `(string)`. It seems this should be consistent.

Guite: Spacing for casts were not consistent before already. Improved that in 7a1aa6ab9130a1ffde0c5c997d5fe5c005ea7cac

src/Zikula/CoreBundle/Command/GenerateTestUsersCommand.php
Guite: Spacing for casts were not consistent before already. Improved that in 7a1aa6ab9130a1ffde0c5c997d5fe5c005ea7cac

Guite:
> why not add this to our own composer file?

I tried that before but it seems to be ignored; _probably_ because `symfony/string` is already covered by `symfony/symfony`. Hence, I referenced #4352

src/Zikula/CoreBundle/Composer/MetaData.php
Guite: The Stringable interface is introduced with PHP 8. Like shown in https://symfony.com/blog/new-in-symfony-5-1-string-improvements#use-stringable-interface the Polyfill provides it already.

> This will allow you to use union types such as string|Stringable in the future when you later upgrade to PHP 8.

At the moment our methods expect `string` for both return types and arguments. So to avoid BC breaks I added the casts in the first step. Changed them to calling `__toString()` now (a3e4402784b5bc3c34cdda98a1859de59690116c).

src/Zikula/CoreBundle/Composer/Scanner.php
Guite: Added `__toString()` call in bd77d26e3d554ba2894d9e53b90ebe3ad97b84b8

craigh:
can use use StyleCI to enforce future casts to require the space?

src/Zikula/CoreBundle/Composer/MetaData.php
craigh: Please add a ticket to 4.0 to change type hints to `\Stringable` where possible?

craigh:
> I tried that before but it seems to be ignored; probably because symfony/string is already covered by symfony/symfony. Hence, I referenced #4352

I mean add the files requirement directly to the composer file

Guite:
> can use use StyleCI to enforce future casts to require the space?

Done in 32f350ecb4018612058870e1f4511d3d71c9669d

src/Zikula/CoreBundle/Composer/MetaData.php
Guite: Follow-up created in #4516

Guite:
> I mean add the files requirement directly to the composer file

Done in 264e51dbeaf353547e53a44a736c9d9e463a5d41

Guite:
>> I mean add the files requirement directly to the composer file

> Done in 264e51d

Needs to be changed in distribution, too.

src/system/RoutesModule/Helper/SanitizeHelper.php
```diff
@@ -13,40 +13,34 @@
...
     public function sanitizeAction(string $methodName): array
     {
-        if ('Action' !== substr($methodName, -6)) {
-            $methodName .= 'Action';
-        }
-
-        $methodName = ucfirst($methodName);
-        $func = lcfirst(substr($methodName, 0, -6));
+        $methodName = ucfirst(s($methodName)->ensureEnd(self::SUFFIX_ACTION)->toString());
+        $func = lcfirst(s($methodName)->trimEnd(self::SUFFIX_ACTION)->toString());
```
craigh, line 43: didn't you just remove `Action` from most (all) of our controller methods? should this be changed?

Guite: I was unsure and decided to review/check this independently as it would include another concern into this PR which is not about changing any behaviour.

craigh: seems a good approach as long as it is on your radar. Another thought with that is the main controller - selecting controller/action for the main page in general settings-> start page... similar issue in Routes module for creating custom routes. also related is #4376.

craigh (approved):

src/system/RoutesModule/Helper/SanitizeHelper.php
Guite: Collected these things in #4517

PR #4558, opened 2020-12-05
replace component installer 

craigh:
replace component installer with oomphinc/composer-installers-extender
remove RequireJs config generator (ping @Kaik)
remove forced use of composer 1 and allow composer 2 in travis build and github actions

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes


craigh:
@Guite @Kaik This removes RequireJs stuff. Since I don't know what it was doing in the first place, I don't know why it was there. It was using classes from component installer (and _**Assetic**_ !!) so if we need it, then we have to come up with a better way. please let me know.

Guite (approved):

Guite:
Did you also consider https://github.com/fxpio/foxy ? (see "Alternatives" section at #4032 ...)

craigh:
> Did you also consider https://github.com/fxpio/foxy 

yes, I looked into it. oomphinc/composer-installers-extender is a smaller change and easier for the moment. Foxy could be used later maybe.

Kaik:
Hard to say now why RequireJs was required by component installer, or why it generated config for it. AFAIR it wasn't even configured properly.
SF and we use Webpack now.

Kaik:
ah ok so, it was an alternative way to handle component js/css we used the simple way (we just hardcoded all assets). These were also "bundled" via requireJS. So instead of many hardcoded tags you use one tag the same everywhere and in your js code you can just call required "module" (JQuery etc...) and requireJS will provide it. 
Webpack does the same (and more) but again we are not using it this way (actually looks like we are not using Webpack Encore at all)
Anyway requireJS can be safely removed.

Guite:
Yes, requireJS is not needed.
Encore becomes more interesting now that Stimulus is going to provide a Flex hook so bundles can be incorporated into Webpack.

Issue #3601, opened 2017-06-13
[Theme] consider webpack-encore


craigh:
http://symfony.com/blog/introducing-webpack-encore-for-asset-management

https://github.com/symfony/webpack-encore

http://symfony.com/doc/current/frontend.html

Guite:
Slightly related: http://symfony.com/blog/new-in-symfony-3-3-manifest-based-asset-versioning

Guite:
https://medium.com/@nioperas06/integrate-webpack-in-your-symfony-application-with-webpack-encore-ed338298a031

Guite:
https://medium.com/a-young-devoloper/symfony-encore-an-awesome-library-4c7da3301139

Guite:
https://github.com/symfony/webpack-encore/issues/5

Guite:
http://symfony.com/blog/new-in-symfony-3-4-http-immutable-responses

Kaik:
This is straightforward we do not have to refactor anything. 
Currently, our js in the core is so poor it won't be affected at all...
Features that webpack encore provides are far beyond what core currently needs/uses. 

After installing we will have assets dir in our src folder where we can place all core assets sources.
Then on deploy, all assets will be compiled sass minified if needed and placed in web/build dir.
We could use it in "legacy" way.
```
window.Twig = require('twig');
window._ = require('underscore');
```
This way we can include jquery as well and all needed libraries.
It does have it's own "composer.json" and ''composer.lock" files so we can save our current versions of assets this way we will be able to manage all needed js css libraries without the need for all those "proxy" bundles...

Kaik:
Hi so I have played a bit with it and I feel safer now in this matter so first what we need.
We need node.js - and yarn or npm as package managers.
Node.js can be installed in various ways and I recommend to install it via nvm - node version manager.
Then depending on how deeply we want to integrate we can:
- install only js side so encore via yarn webpack etc... and include files that are created in our assets listener.
-  install php side and this can be done via composer or flex (at the moment there is one conflict)

Core assets can be declared directly src/assets/js and /css. All assets can be managed and processed by loaders.
We could completely move all asset packages from php side "composer" to webpack.

Php bundle which needs to be installed, enabled and then it will provide additional integration like twig etc.. I have not tested this yet.

I'm not sure how to handle modules and theme assets yet including bootstrap etc..

Guite:
> We could completely move all asset packages from php side "composer" to webpack.

[...]

Issue #3438, opened 2017-02-04
[Extensions] Install extensions using composer


Guite:
**Depends on #3644**

Use composer for installing extensions during runtime.

We could also use our extension library as a composer repository then and/or use packagist.
Thus this would also solve #909
See also https://github.com/composer/satis

Example: [Contao composer client](https://github.com/contao-community-alliance/composer-client) ([German manual](http://de.contaowiki.org/Composer/Client)).
Example: [Drupal composer manager](https://www.drupal.org/project/composer_manager)

- [ ] POC for PHP-driven runtime executions
  - [x] Evaluations
     - [ ] [Composer PHP docs](https://github.com/composer/composer/tree/master/doc/)
        - [ ] https://github.com/composer/composer/blob/master/doc/03-cli.md#depends
        - [ ] https://github.com/composer/composer/blob/master/doc/03-cli.md#prohibits
     - [ ] [Composer PHP API](https://getcomposer.org/apidoc/master/index.html)
     - [ ] [PageKit example](https://github.com/pagekit/pagekit/blob/develop/app/installer/src/Helper/Composer.php)
  - [x] Preparation: Read core dependencies out of `composer folder / installed.json`
  - [ ] Module installations and updates
     - [ ] Read core dependencies and merge with module dependencies
     - [ ] Install without changing core dep versions
     - [ ] If core dep versions would be changed the module can not be installed
  - [ ] Core updates
     - [ ] Set all modules to inactive whose dependencies conflict with updated core dependencies
- [ ] Tests
  - [ ] Install module with dependency to Symfony bundle
  - [ ] Uninstall module with dependency to Symfony bundle
  - [ ] Install module with dependency to other module
  - [ ] Uninstall module with dependency to other module
  - [ ] Conflict
- [ ] Further aspects
  - [ ] Rewrite our current extensions management.
  - [ ] Unify handling of modules and themes.
  - [ ] Also think about handling translations.
  - [ ] Handle legacy components

rallek:
can we use this to add also language packs? Including language packs for modules and themes?

craigh:
- https://www.drupal.org/node/2718229
- https://packagist.drupal-composer.org/packages/drupal/channel
- https://packagist.drupal-composer.org/explore/

discordier:
As @Guite contacted me via mail on this issue for feedback how we did this for Contao, I'll add some relevant links here.

[...]

PR #4563, opened 2020-12-12
Add symfony/flex dependency

craigh:
This removes `zikula/core` dependency on `symfony/symfony` and instead requires needed components. Additionally, it adds dependency on `symfony/flex`. Flex is not useful in this state however except for core development (which seems minimal). Once this is merged, the distribution would need to be updated as well to ensure the flex dependency is used there. When this is done, real-world (downline) development starting with the distribution becomes possible using flex.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | #4352, #3601, #3438, #3962
| License           | MIT
| Changelog updated | yes

Using flex has become increasingly needed with the premier of the `symfony/ux` initiative which relies on webpack-encore as well. I think we could implement webpack-encore without using it to "install" our primary front-end dependencies (bootstrap, jquery, etc). but instead use it initially only for the `core.css` file (and maybe the jQuery config file) and then, allow it to be used for downline development.


Guite (dismissed):
I really like this change.
But I vote for adding all Symfony components for 3.x to ensure BC.

Background: until #3438  is done the core can not ensure that no conflicts may occur because of extensions requiring additional Symfony components (in possibly too old or too new version). Hence, extensions need to rely on the core providing the entire Symfony ecosysystem. Of course this has limits (e.g. we should not include all possible notifier packages), but the most important components need to be kept available even if the core does not use them internally. One example for this is the messenger component.

craigh:
@Guite, I have added the remaining components AFAIK except those noted in the updated changelog. Not noted there, these are also not added:
 - symfony/inflector (deprecated)
 - symfony/semaphore (experimental)
 - symfony/uid (experimental)

I believe semaphore and uid were added in 5.2 anyway, so there is no reason to include them. inflector is to be replaced by the string component.

I've worked hard to ensure BC. I believe I've got all as needed. Please review again.

Guite (approved):
Nice!

Adding uid soon will become useful though, but it is not urgent :smile:

craigh:
@Guite any further thoughts or concerns here? If not, I will merge this later when I have time. Looks like I will do it locally to deal with the conflicts.

Guite:
No objections.

PR #4571, opened 2020-12-25
add Symfony/WebpackEncoreBundle

### LINKED ISSUE #3601: [Theme] consider webpack-encore

craigh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | fixes #3601
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes

I'm not sure how best to proceed with providing _final_ builds to end users through the distribution. In a normal situation, we would not commit the contents of `/public/build` to github. We could maybe do this in the distribution but this could be problematic if the user installs in a subdirectory.

currently the **publicPath** value in `webpack.config.js` is hard-coded to MY value `/core.git/public/build` - it would default to simply `/build` - because I work in a subdir all the time.

Adding this complicates the installation process but provides later flexibility to downline developers.

Additional things I thought of:
 - on install, need to confirm if installing in subdir and adjust `setPublicPath('/build')` in `webpack.config.js`
     - (must also `setManifestKeyPrefix('build/')` be enabled?)
     - also this path may be added to .gitignore (`/public/build/`)
 - can the installer run `yarn encore dev` ?


craigh:
I guess one option would be to commit the contents of `/public/build` with default values. Then on install, do a simple string replace if the install is in a subdirectory. 

Doing a search, it looks like there are instances of the path in
 - `/public/build/manifest.json`
 - `/public/build/entrypoints.json`
 - `/public/build/runtime.js`

opinions?

craigh:
@Guite I think I have this ready now. It includes the default assets from webpack and on install, adjust the path if needed.

Guite (approved):
Terrific :smile:

PR #4572, opened 2020-12-31
use addAnnotatedClassesToCompile

craigh:
Bundles can hint Symfony about which of their classes contain annotations so they are compiled when generating the application cache to improve the overall performance. Define the list of annotated classes to compile in the addAnnotatedClassesToCompile() method:

https://symfony.com/doc/current/bundles/extension.html#adding-classes-to-compile

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Deprecations?     | no
| Fixed tickets     | -
| Refs tickets      | -
| License           | MIT
| Changelog updated | yes


Guite (approved):

Guite:
How about something like `'**Module\\Controller\\'` to include 3rd party modules, too?

craigh:
> How about something like '**Module\\Controller\\' 

I did consider it, but elected to just include core for now. Any module can include it on their own of course.

